Issue #11, opened 2022-12-23
Support for `q.literal`


gksander:
We should add a `q.literal` utility that maps to Zod's `z.literal` so that you can query with validation for a literal type.

Issue #10, opened 2022-12-23
Add documentation for `coalesce` queries


gksander:
As it stands, I believe you can do `coalesce` in queries via something like:

```ts
q.grab({
  title: [`coalesce(${field}.${language}, ${field}.en, '')`, q.string()]
})
```

We should write some tests to validate this, and then document the behavior.

PR #12, opened 2022-12-23
Literal type

gksander:
Address #10 and #11 : 

- Adds support for a `q.literal` type (that just wraps Zod's `z.literal`).
- (bonus points) Adds support for `q.union` type (that just wraps Zod's `z.union`).
- Adds tests and README deets on using groq's `coalesce` operator.


scottrippey (approved):
Looks good, nice tests!

Issue #21, opened 2023-01-09
Alternative API Suggestion: Builder Pattern


nonphoto:
First off thanks for creating this library! This is something that I have been really needing and I think the approach of typing the queries rather than Sanity schemas is the right way to go given the unstructured nature of the Sanity content lake.

In an effort to improve what's already been started I want to suggest a change to the API that I think could simplify a few things, ensure type safety in more places, and even open up the option to implement other groq features.

### Issues with the current API

Here are a few issues that I ran into while using this library.

#### Lack of type safety for sequential grabs

Pipes don't have any awareness of the type they are operating on. This is fine if the user hasn't defined the type for the data they are querying, but it can cause problems if they have. Consider the following:

```
q(
  "*",
  q.filter("_type == 'post'"),
  q.grab({
    title: q.string(),
  }),
  q.grab({
    date: q.date(),
  })
)
```

The above should not be allowed because the second grab will never have a date field to grab. This result of `schema.parse` is `[{}, {}, ... ]`

#### Lack of type safety for filters

Similar to the above, the parameter for `filter` is a string and can't enforce that it's operating on a query result of the correct type.
For example:

```
q(
  "*",
  q.grab({
    title: q.number()
  }),
  q.filter("title == 'hello'")
)
```

The above should not be allowed because title is a number, not a string.

#### Lack of type safety for slice

```
q("*", q.slice(0), q.filter("_type == 'menu'"))

**Comment by gksander** [comment]
Wow, thank you for this very thoughtful contribution! 

I'm a fan of the builder pattern, especially when it comes to TS and typings. I think I played around with a builder pattern similar to what you proposed here when doing early exploratory work with groqd, and I don't quite recall why I ended up not going with the builder pattern. I don't think it was a technical limitation.

I'll definitely give some thought to this and do some more exploration. I agree that a builder pattern could help make this library more robust (and likely easier to maintain!) 

I've got a busy week with some other obligations, but will hopefully be able to free up some time to dig into this here soon!

**Comment by nonphoto** [comment]
Wow thanks for incorporating this suggestion so quickly! I have some thoughts on additional features, but this is already more than enough for my needs.

For sequential grabs and filters, I think the right move would be to extend the DSL to include GROQ functions:

```
*[]{ name }{ "foo": coalesce(name, "no name") }

nonphoto:
Wow thanks for incorporating this suggestion so quickly! I have some thoughts on additional features, but this is already more than enough for my needs.

[...]

PR #22, opened 2023-01-10
Builder Pattern

gksander:
**A big one...**

This PR was motivated by #21, and provides a _large_ refactor of the library to use a Builder Pattern instead of a pipeline pattern. This provides a good handful of benefits:

- Much easier to follow what's going on with the library, and should make it much easier to maintain and contribute to.
- Allows us to better restrict what operations can be used when. For example, doesn't make sense to order/slice/filter a single entity, so with the builder pattern – we can better enforce that.
- Makes the end-user TS experience much better and less convoluted.
- Also addresses some other outstanding issues I had been mulling on.


src/builder.test.ts
```diff
@@ -0,0 +1,398 @@
+import { describe, it, expect } from "vitest";
```
gksander, line 1: This file is just a conglomerate of all the old test files that got deleted, just tweaked to use the new builder pattern.

README.md
```diff
@@ -45,34 +41,50 @@ export const runQuery = makeSafeQueryRunner(query => client.fetch(query));
...
+    .grab({ name: q.string() })
+    .slice(0, 150)
 );
 // data: { name: string }[]
 ```
 
 Using `makeSafeQueryRunner` is totally optional; you might find using `q().schema` and `q().query` in your own abstractions works better for you.
 
-## Why? 🤷‍
+## **NOTE**: Significant API changes with 0.3.0
```
gksander, line 54: Ugh, I'm torn on this. These are breaking changes, and should probably be released as a major version change. Maybe I should be planning to release as a 1.0.0-alpha.

fritz-c: In sub-v1.0.0 semantic versioning, 0.3.0 is appropriate for breaking changes.

image
https://semver.npmjs.com/

src/grab.ts
```diff
@@ -1,157 +1,123 @@
...
-        return prev.schema instanceof z.ZodArray
-          ? z.array(prev.schema.element.pick(toPick))
-          : (prev.schema as z.ZodObject<any>).pick(toPick);
-      }
-      // If not unknown/object, I don't know what happened 👀
-      else {
-        return z.never();
-      }
-    })();
+  const res = (schema instanceof z.ZodArray
```
fritz-c, line 95:
```suggestion
  const res = (newSchema instanceof z.ZodArray
```
Is this the answer to your TS problem below?

src/builder.test.ts
```diff
@@ -0,0 +1,398 @@
...
+    );
+    invariant(data);
+    expect(data[0].hp).toBe(45);
+  });
+
+  it("can grab with {key: q()} composition", async () => {
+    const { query, data } = await runPokemonQuery(
+      q("*")
+        .filter("_type == 'poketype'")
+        .slice(0, 1)
```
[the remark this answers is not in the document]
gksander: Ahh that's an interesting point! I'm using `slice` as a name here, because that's the Groq terminology for this operation.

src/builder.test.ts
```diff
@@ -0,0 +1,398 @@
...
+        .grab({
+          name: z.string(),
+          pokemons: q("*")
```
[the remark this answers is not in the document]
gksander: Man, I'm such a noob!

src/builder.test.ts
```diff
@@ -0,0 +1,398 @@
...
+    );
+    invariant(data);
+    expect(data[0].name).toBe("Chansey");
+  });
+});
+
+describe("ArrayResult.slice", () => {
+  it("turns unknown[] to unknown if no max provided", async () => {
+    const { query, schema, data } = await runPokemonQuery(
+      q("*").filter("_type == 'pokemon'").slice(0)
```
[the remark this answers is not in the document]
gksander: I like this thought – let me file this as an issue so I can give it a little more thought, and to keep this PR from being too wild.

src/builder.test.ts
```diff
@@ -0,0 +1,398 @@
...
+
+    invariant(data);
+    expect(data[0]).toEqual({ _id: "pokemon.1", name: "Bulbasaur" });
+    expect(data[1]).toEqual({ _id: "pokemon.2" });
+    expect(data[3]).toEqual({ _id: "pokemon.4", name: "Charmander", hp: 39 });
+
+    for (const dat of data) {
+      if ("name" in dat && dat.name === "Charmander") {
+        expect(dat.name === "Charmander").toBeTruthy();
+        // @ts-expect-error Expect error here, TS should infer type
```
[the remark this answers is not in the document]
gksander: 😅 I won't try to rank these cases, instead I'll just add your suggestion alongside what I have here. Thanks!

README.md
```diff
@@ -86,69 +98,61 @@ const client = sanityClient({ /* ... */});
...
-    //  projection is `{ "types": types[]->{ name } }`
-    types: q("types", q.filter(), q.deref(), q.grab({ name: q.string() }))
-  }),
-);
+q("*").grab({
+  // projection is `{ "name": name }`, and validates that `name` is a string.
+  name: ['name', q.string()],
+
+  // shorthand for `description: ['description', q.string()]`,
+  //  projection is just `{ description }`
```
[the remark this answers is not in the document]
gksander: groq offers a shorthand where instead of writing `{ "name": name }` you can write `{ name }`. In my experience, I see most people prefer the shorthand – unless they need to rename a field or do something more complex than just grab a field.

I'm not sure there'd be much benefit in expanding `{ description }` out to `{ "description": description }`, it's mostly just added bytes. Maybe the example `name: ['name', q.string()],` is just a bit too simplified. The reason that first form is an option is so you can do something like:

```ts
q("*").grab({ name: ["coalesce(name, 'no name')", q.string()] });
```

or something of the likes.

src/builder.test.ts
```diff
@@ -0,0 +1,398 @@
...
+  it("applies order, preserves unknown schema", async () => {
+    const { query, schema, data } = await runPokemonQuery(
+      q("*").filter("_type == 'pokemon'").order("name asc")
+    );
+
+    expect(query).toBe(`*[_type == 'pokemon']|order(name asc)`);
+    expect(schema instanceof z.ZodArray).toBeTruthy();
+    expect(schema.element instanceof z.ZodUnknown).toBeTruthy();
+    // @ts-expect-error data is unknown type since no grab present
+    expect(data?.[0]?.name).toBe("Abra");
```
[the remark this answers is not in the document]
gksander: hmm, I'm not sure I follow. Right now `data` has a type of `unknown[]` since no `grab` call was made. My line of thinking here is: TS doesn't have a way to know that `data` is an array of objects that have a `name: string` field, so we'd expect TS to balk. I'm not sure we'd want to force a type on `data`, because the goal of this test case is to show that there is no (useful) type without a `grab` or `grabOne` call.

README.md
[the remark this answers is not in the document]
fritz-c: I see. So it's not so much shorthand within this library (which produces different GROQ for each syntax) as it is shorthand upon being used in GROQ.

src/grab.ts
gksander: WOAH YEAH. Great catch here! The `schema instanceof z.ZodArray` was there to indicate that if the _previous_ schema was an array, the new result should also be an array. But the `newSchema` generation logic already has that logic baked in, so we can use `newSchema` in this check.

src/builder.test.ts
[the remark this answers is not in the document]
gksander: Ahhh gotcha! I see what you're saying. I hadn't thought much about that, because bypassing the Zod validation (by skipping the `.grab` calls) eliminates a lot of the utility of the library. But I'd say your best bet is just casting, exactly as you have shown above.

README.md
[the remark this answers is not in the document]
gksander: Ummm I think it might be _both_ – a shorthand used within this library (because `name: q.string()` is nicer to write than `name: ["name", q.string()]`) that also maps nicely into an official groq shorthand (using `{ name }` instead of `{ "name": name }` for the projection).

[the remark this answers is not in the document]
fritz-c: Makes sense 👍

src/grab.ts
```diff
@@ -1,157 +1,121 @@
 import { z } from "zod";
-import { BaseResult, ValueOf } from "./types";
+import { ValueOf } from "./types";
+import { ArrayResult, BaseResult, EntityResult } from "./builder";
```
fritz-c, line 3: Regarding the class names, `*Result` was a bit confusing to me because it seems like it would be the actual output stored inside, rather than just the query and anticipated schema. Would you be open to alternate naming? The best I could come up with was `*Intent` or `*Plan`, though those may just trade in the misleading aspect for plain old confusion ("what's an `EntityIntent`?").

Perhaps there is precedent in other query builders that we could reference? I imagine many might call it `*Query`, which wouldn't be too bad either. Just a smidge confusing in that that class is more than just the query string.

gksander: 100%, I totally agree. I had that thought pop in my head at some point, but it must have gotten lost in the thick of things. Thanks for bringing it back up – I think I like `*Query`, that seems to make sense to me!

src/builder.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.query += `[${filterValue}]`;
+    return new UnknownArrayResult({
+      ...this.value(),
+      schema: z.array(z.unknown()),
+    });
+  }
+
+  grab<
+    S extends Selection,
+    CondSelections extends Record<string, Selection> | undefined
```
fritz-c, line 41: Just curious - is there a benefit to naming the types up in the generic parameter area versus just putting them in the function argument area below?

src/builder.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  public value(): Payload<T> {
+    return { schema: this.schema, query: this.query };
+  }
+}
+
+/**
+ * Unknown, comes out of pipe and is starting point for queries.
+ */
+export class UnknownResult extends BaseResult<z.ZodUnknown> {
```
fritz-c, line 25: Not 100% sure it's a good idea, but after looking at your classes, it appears you could refactor and define `UnknownResult` by extending `EntityResult`, in the same way `UnknownArrayResult` extends `ArrayResult`.

src/grab.ts
```diff
@@ -1,157 +1,121 @@
...
 
+/**
+ * Misc util
+ */
 type Field<T extends z.ZodType> = T;
 type FromField<T> = T extends Field<infer R>
   ? R
   : T extends [string, infer R]
   ? R
   : z.ZodNever;
```
fritz-c, line 116: Any reason you use `z.ZodNever` instead of `never`? My understanding of it is that if you've reached this point in the nested ternary, it's an issue with what the dev is passing as a Selector; it doesn't conform to the groqd-accepted syntax of `q.boolean()` or `["active", q.boolean()]`, for example. So the problem is more in groqd-land than it is in zod-land.

src/grab.ts
```diff
@@ -1,157 +1,121 @@
...
+  T extends z.ZodTypeAny | z.ZodArray<any>,
+  S extends Selection,
+  CondSelections extends Record<string, Selection> | undefined
+>(
+  query: string,
+  schema: T,
+  selection: S,
+  conditionalSelections?: CondSelections
+) => {
+  type FromSelection<Sel extends Selection> = z.ZodObject<{
```
fritz-c, line 15: This type could be moved down with your other utility types, as it doesn't have any dependencies inside this function.

src/builder.ts
gksander: Another great catch! That doesn't seem to choke any tests, and reduces even more code. Dope!

src/builder.ts
gksander: I've got those in generics because I have conditional types like this:

```ts
type NewType = T extends z.ZodArray<any>
    ? ArrayQuery<AllSelection>
    : EntityQuery<AllSelection>;
```

that depend on them in the function body. The function signature varies as a function of these generic types, which is why I put them in the generics section instead of in the function argument itself.

gksander: Yeah, the fact that these are generic types is part of what powers the type-inference of the whole thing. If you just specify a type for the function arguments, then you won't get the strict type inference.

It's a bit hard to describe in writing, but if you try moving these from generics into the function arguments, and then open the `builder.test.ts` or `index.test.ts` files and start to view some definitions for various `data` variables, you'll notice you lose out on the inference.

fritz-c: Ah, I think I get it. You pass the full type (the thing extending your base types) down to the next function rather than it just resetting to the base types you specify in the function arguments. Lots of mind-melting TS concerns at this level of handing types. 🔥🤔🔥

src/grab.ts
```diff
@@ -1,157 +1,121 @@
...
+    const baseSchema = z.object(toSchemaInput(selection));
+    const conditionalFieldSchemas = conditionalFields.map((field) =>
+      baseSchema.merge(z.object(toSchemaInput(field)))
+    );
+    const s =
+      conditionalFieldSchemas.length === 0
+        ? baseSchema
+        : // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore Need to figure out how to make tsc happy
+          z.union([...conditionalFieldSchemas, baseSchema]);
```
fritz-c, line 90:
```suggestion
        : z.union([
            baseSchema,
            conditionalFieldSchemas[0],
            ...conditionalFieldSchemas.slice(1),
          ]);
```

src/grab.ts
```diff
@@ -1,157 +1,121 @@
...
-        }, {});
+  const res = (newSchema instanceof z.ZodArray
+    ? new ArrayQuery({
+        query: query + `{${projections.join(", ")}}`,
+        schema: newSchema,
+      })
+    : new EntityQuery({
+        query: query + `{${projections.join(", ")}}`,
+        schema: newSchema,
+      })) as unknown as ArrayQuery<AllSelection>;
```
fritz-c, line 97:
```suggestion
      })) as unknown as NewType;
```
Then you can omit the cast on the return statement.

fritz-c (approved):

PR #23, opened 2023-01-10
Add the Dark-type and Tyranitar to sample data.

carlos-kelly:
We're missing the Dark-type in our sample data. Also Tyranitar is cool.


gksander (approved):
🦖

Issue #26, opened 2023-01-16
Deref without array and nullable queries


olemagnus:
I upgraded to version `0.3` and I'm having some issues recreating a query. `deref()` used to work without `filter()`. I have a reference field that might be empty, and there doesn't seem to be a way to use `nullable()` on on `q()` to handle if there are no references. 

This code worked with `0.2`:
```
role: q(
  "role",
  q.deref(),
  q.grab({
    title: q.string(),
  })
)
```

With `0.3` I get this error: "Property 'deref' does not exist on type 'UnknownQuery'.":
```
role: q("role")
  .deref()
  .grab({ title: q.string() })
```

Adding `filter()` gives me a runtime error because the list can be empty (null):
```
role: q("role").filter().deref().grabOne("title", q.string()),
```

gksander:
Thank you for reporting! This appears to be a regression, but should be an easy fix! I'll try to sneak in a fix here today or early tomorrow.

gksander:
Closing as resolved via #27, released as 0.3.1. Let me know if you're still having issues with this after upgrading to 0.3.1!

PR #27, opened 2023-01-16
`.deref` on `UnknownQuery`

gksander:
Allow for `deref` on `UnknownQuery`, addresses #26.

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -61,6 +61,11 @@ export class UnknownQuery extends EntityQuery<z.ZodUnknown> {
       schema: z.array(z.unknown()),
     });
   }
+
+  deref() {
+    this.query += "->";
+    return this;
+  }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/builder.ts
```diff
@@ -61,6 +61,11 @@ export class UnknownQuery extends EntityQuery<z.ZodUnknown> {
       schema: z.array(z.unknown()),
     });
   }
+
+  deref() {
+    this.query += "->";
```
carlos-kelly, line 66: do we need to do any enforcement that this should be the last in the chain?

gksander: Generally, a `.grab` call will be last in the chain, and having `deref` as last in the chain would leave you with an `unknown` result. So having `deref` last in the chain doesn't make much sense from a TS perspective, but groq doesn't care if you do it.

carlos-kelly (approved):

PR #30, opened 2023-01-21
Support for image queries

gksander:
This PR contains a Sanity image helper, which is purely a convenience method (which could be achieved in user-land with raw groqd queries). The goal of this helper is to help make user-land code more concise for common image scenarios.

There's a lot to this PR (I had to make some changes to get everything working as expected), but the best approach is probably:

- [ ] Read the changes to `README.md` to see what the API looks like.
- [ ] Then read the tests for `sanityImage.test.ts` to see what the API looks like in action, and behavior we're expecting.
- [ ] Then suffer your way through `sanityImage.ts` to see how the sausage is made.
- [ ] Rest of the changes are just auxilary to support the additions made in `sanityImage.tes`.


carlos-kelly (approved):

src/sanityImage.test.ts
```diff
@@ -0,0 +1,294 @@
...
+          }),
+        })
+    );
+
+    expect(query).toBe(
+      `*[_type == 'pokemon'][0..1]{name, "cover": cover{_key, _type, "asset": asset->{"metadata": metadata{"location": location{_type, lat, lng}}}}}`
+    );
+    const location = data?.[0]?.cover?.asset?.metadata?.location;
+    invariant(location);
+    expect(location._type === "geopoint").toBeTruthy();
```
fritz-c, line 218:
```suggestion
    expect(location._type).toEqual("geopoint");
```
Maybe restructure the expect calls to be like this, for better error messages.

fritz-c (approved):

src/sanityImage.test.ts
gksander: I've been dying to have a conversation with someone about this! The test suite here looks _different_ than how I'd typically write unit tests, primarily to test the TS types! The unit tests serve a dual purpose: runtime behavior, and TS typings.

In this scenario, `location._type` should be the exact literal `"geopoint"` – and it's hard to capture that sort of type-checking with tests like `expect(location._type).toEqual("geopoint")`, since that latter example doesn't make any type-level assertions about what the type/literal value of `_type` is.

I'm still trying to flesh my thoughts out regarding this, but that's a little insight into why these tests look the way they do.

fritz-c: That makes sense. It's fine as-is, then.

PR #39, opened 2023-01-26
`score` function

gksander:
Adds `score` function support. Somewhat self-explanatory, behaves very similar to the `order` method.


carlos-kelly (approved):
Once again, I learned what score did by noticing the Charmander line before the others. This is pretty cool and I could see some practical uses for it on our projects.

Issue #32, opened 2023-01-25
Feature: Block Content helper


gksander:
Like with `q.sanityImage`, we should consider adding a `q.blockContent` helper for block content fields.

gksander:
Closed via #42

PR #42, opened 2023-01-27
`q.contentBlock` helper

gksander:
Adding a little schema for `q.contentBlock` to make it easier to fetch content blocks from Sanity, useful with Sanity's WYSIWYG block editor. Addresses #32


carlos-kelly (approved):
This is great and a pretty common field!

Issue #31, opened 2023-01-23
Example: Params in query


heggemsnes:
Would be great with an example on how to use params like $type or $id in the queries. I hacked something together but there's probably a better way.

heggemsnes:
Yes that works nicely for strings, but can cause issues if you are trying to query by a number i.e.

``q("*").filter(`_type == "pokemon" && base.Attack > ${attack}`).slice(0)``

Doesn't really work where:

``q("*").filter(`_type == "pokemon" && base.Attack > $attack`).slice(0)``

Does work.

Also useful for arrays,  like this:

```
const docIds = ["document_id_1", "document_id_2"];
q("*").filter(`_type == "article" && references($docIds)`).slice(0)
```

I copied some types and added this to do this, would be easier if these types are exported :)

```
export const makeSafeQueryRunnerCustom =
  (fn: QueryExecutor) =>
  <T extends BaseType>(
    client: SanityClient,
    {query, schema}: BaseQuery<T>,
    params?: Object
  ): Promise<z.infer<T>> =>
    fn(client, query, params).then((res) => schema.parse(res))

export const tClient = makeSafeQueryRunnerCustom((client, query, params) =>
  client.fetch(query, params ?? {})
)

```

gksander:
Oh wow, you're absolutely right! Wonderful callout. That should be an easy add - I can probably get that added in today!

gksander:
Updated via #43, and released with 0.6.0! Thanks for calling this out, this is a great addition to the library! I tested on my end in a project, but let me know if you give it a try and it doesn't quite work as you'd expect!

PR #43, opened 2023-01-27
Add flexibility to `makeSafeQueryRunner`

gksander:
Addresses #31 and adds some flexibility to `makeSafeQueryRunner` so that you can e.g. pass parameters via `sanity/client`'s params argument.


carlos-kelly (approved):
Awesome. I could see this being useful if you need to pass headers or something else.

PR #54, opened 2023-02-01
Extend InferType to accept zod schema

gksander:
Allow `InferType` to accept Zod schema, for use-cases where you're destructuring `schema` out of your query:

```ts
import type { InferType } from "groqd";
import { q } from "groqd";

const query = q("*").filter().grab({ name: q.string() });
type Thing = InferType<typeof query>;

// OR!

const { schema } = q("*").filter().grab({ name: q.string() });
type Thing = InferType<typeof schema>;
```

### DIFF
--- a/.changeset/cuddly-meals-tease.md
+++ b/.changeset/cuddly-meals-tease.md
@@ -1,0 +1,5 @@
+---
+"groqd": patch
+---
+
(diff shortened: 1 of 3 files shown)


umxr (approved):
LGTM! this is super helpful :)

PR #63, opened 2023-02-16
`TypeFromSelection` Utility Type

gksander:
This type will be handy-dandy, especially when extracting sub-queries/selections out into their own files and using their type in e.g. a React component's props. (Had to update some deps for TS `satisfies` operator.)


gksander:
Whoops, looks like upgrading TS broke something.

src/builder.test.ts
```diff
@@ -253,6 +253,7 @@ describe("ArrayResult.slice", () => {
     expect(query).toBe(`*[_type == 'pokemon'][0]`);
     expect(schema instanceof z.ZodUnknown);
     invariant(data);
+    // @ts-expect-error data is unknown type since no grab present
     expect("name" in data).toBeTruthy();
```
gksander, line 257: I actually don't know how TS wasn't complaining about this prior to 4.9 🤔

carlos-kelly (approved):
Fantastic, this will be awesome!

PR #67, opened 2023-02-20
feat: added helper for slug fields

### LINKED ISSUE #62: Slug helper
Fetching slugs is very common, and the suggestion right now is to do:

```ts
q().grab({ slug: ["slug.current", q.string() ] })
```

but we might as well ship a `q.slug` for this.

iduuck:
fixes #62

### DIFF
--- a/src/schemas.ts
+++ b/src/schemas.ts
@@ -38,6 +38,8 @@ const contentBlock = () =>
     level: z.number().optional(),
   });
 
+const slug = (fieldName: string) => [`${fieldName}.current`, z.string()];
+
 export const schemas = {
   string: z.string,
   number: z.number,
@@ -50,5 +52,6 @@ export const schemas = {
   union: z.union,
   array: z.array,
   object: z.object,
+  slug,
   contentBlock,
 };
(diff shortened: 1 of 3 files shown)


src/schemas.ts
```diff
@@ -38,6 +38,9 @@ const contentBlock = () =>
     level: z.number().optional(),
   });
 
+const slug = (fieldName: string) =>
+  [`${fieldName}.current`, z.string()] as [string, z.ZodType];
```
gksander, line 42:
```suggestion
  [`${fieldName}.current`, z.string()] as [string, z.ZodString];
```

With `[string, z.ZodType]`, the resulting data type for a slug field is `any` (since `z.ZodType` is broad). If we just tighten this down a tiny bit, we'll get `slug` fields typed as `string` instead (which I think is what we want).

gksander (commented):
One quick suggestion! Also, will you add a changeset to this PR as well via:

```sh
yarn run changeset
```

and we can add this as a minor, which will allow us to ship this (likely) today!

iduuck:
i think we are fine, when the checks are through @gksander

PR #72, opened 2023-02-23
Rage against the (`null`) machine

gksander:
Sanity/GROQ return `null` values for missing values. This becomes annoying to deal with with `groqd` because you end up littering a lot of `.nullable()`s around, but you can't do `.default()` chained on a nullable. This PR adds some helpers to preprocess `null` values to `undefined` so that if a value is missing, we can assume it's `undefined` which allows for more consistent chaining and ability to do things like

```ts
q().grab$({
  nope: q.optional().default("not there")
})
```

and then `data["nope"]` is of type `string`.

---

**Example showing groq returning `null` for a missing value.**

image

### DIFF
--- a/.changeset/lemon-rats-move.md
+++ b/.changeset/lemon-rats-move.md
@@ -1,0 +1,5 @@
+---
+"groqd": minor
+---
+
+nullToUndefined, grab$, and grabOne$ helpers to make dealing with null return values much more friendly.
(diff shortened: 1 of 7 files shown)


carlos-kelly (approved):
This is great, the default value fallback for things that don't even exist is great.

Issue #68, opened 2023-02-20
Type helpers for tests


gksander:
We should add type-level assertions in our tests to ensure output types are as expected. Zod has a useful helper [here](https://github.com/colinhacks/zod/blob/master/src/helpers/util.ts), we should mimic that and decorate our tests with these (and maybe cleanup our tests a bit accordingly, e.g. `expect(data.name).toBe("Bulbasaur")` instead of `expect(data.name === "Bulbasaur").toBeTruthy()`).

PR #75, opened 2023-02-24
fix: fixed conditional selection bug (empty base schema)

### LINKED ISSUE #25: Empty base selection for `grab` should not create an empty object schema for validation
Per [this comment](https://github.com/FormidableLabs/groqd/issues/13#issuecomment-1368138373), we should look into the case where an empty object is provided as the first arg to `grab`. Right now, `grab` will generate an empty object schema that then gets thrown into a union schema, which will cause some weird results. 

In code, we have this:

```ts
// ...
z.union([...conditionalFieldSchemas, baseSchema])
```

and when the first arg to `grab` is an empty object, that `baseSchema` field should be omitted from this union.

iduuck:
fixes #25

also touches #68

### DIFF
--- a/src/builder.test.ts
+++ b/src/builder.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, it } from "vitest";
+import { describe, expect, expectTypeOf, it } from "vitest";
 import { q } from "./index";
 import { z } from "zod";
 import { runPokemonQuery, runUserQuery } from "../test-utils/runQuery";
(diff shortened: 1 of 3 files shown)


iduuck:
Please review, @gksander.

src/builder.test.ts
```diff
@@ -82,6 +82,47 @@ describe("ArrayResult.grab/UnknownResult.grab/EntityResult.grab", () => {
...
+        )
+    );
+
+    expect(query).toBe(
+      `*[_type == 'pokemon'][0..3]{...select(name == 'Charmander' => { name, "hp": base.HP }, name == 'Bulbasaur' => { name, "attack": base.Attack })}`
+    );
+
+    invariant(data);
+    expect(data[0]).toEqual({ name: "Bulbasaur", attack: 49 });
+    expect(data[1]).toEqual({});
```
gksander, line 111: I think this is the thing that is getting me hung up. If there's an empty base selection, and our type looks something like `({ name: "Charmander", ... } | { name: "Bulbasaur", ... })[]`, then having an empty object is a "wrong" according to our type.

But in groq, I believe you can have an empty projection and it'll just return an empty object. So for example, this query:

```
*[_type == 'pokemon'][0..3]{...select(name == 'Charmander' => { name, "hp": base.HP }, name == 'Bulbasaur' => { name, "attack": base.Attack })}
```

does indeed pass the tests you have here, _including_ the empty-object one – since that projection will just select nothing on everything but Charmander and Bulbasaur.

Still trying to wrap my head around this, but I think limiting our types on this might end up causing a mismatch with how groq behaves – which would be confusing.

Any thoughts on this? It's not yet clear to me how to proceed on this.

iduuck: Jep, that's true. Also don't know, how to proceed either. Then, I think, the PR can be closed, and it's up to the user, to check, whether the empty object is matching, or not.

Though, on the other side, if there is an empty object in the response, I don't think there is a big case for any user to show an empty tile/card/list-item/whatever, so my opinion on this would be to just add a third argument to the conditionals, and just set a default option.

I am thinking of an option named `includeEmptyObject`, which is then conditionally giving the user at least the type. But then, we would need to filter the empty object on our side.

littlemilkstudio:
So we're working on an idea with a new `select()` method that was part of the `0.10.0` release. You can check out the details [in this PR here](https://github.com/FormidableLabs/groqd/pull/82), but it's basically for the case where you have an empty base selection, and your sanity data falls into one of a few shapes.

To omit the empty object (or null depending on which context you use it), you pass it a default case. So here it would look like:
```ts
const { data, query } = await runPokemonQuery(
      q("*")
        .filter("_type == 'pokemon'")
        .slice(0, 3)
        .select(
          {
            "name == 'Charmander'": {
              name: q.literal("Charmander"),
              hp: ["base.HP", q.number()],
            },
            default: {
              name: q.string(),
              attack: ["base.Attack", q.number()],
            },
          }
        )
    );
```
or if you wanted specifically Bulbasaur & Charmander, you could add it to the filter, ie:
```ts
const { data, query } = await runPokemonQuery(
      q("*")
        .filter("_type == 'pokemon' && (name == 'Charmander' || name == 'Bulbasaur')")
        .select(
          {
            "name == 'Charmander'": {
              name: q.literal("Charmander"),
              hp: ["base.HP", q.number()],
            },
            default: {
              name: q.literal("Bulbasaur"),
              attack: ["base.Attack", q.number()],
            },
          }
        )
    );
```

scottrippey:
In the `groq-builder` branch, we've addressed this issue by automatically including the empty object in the result type, but there's an option called `isExhaustive` that won't include the empty object type.

scottrippey:
`groq-builder` is now merged and released as `groqd @ 1.0`
With this version, in `conditional` and `conditionalByType` there's an `isExhaustive` option, which will omit the empty object.  
Even better, with `conditionalByType`, the exhaustive check is automatic -- it detects if you've covered all possible types.

PR #79, opened 2023-02-28
Bug: Conditional Selection Type

littlemilkstudio:
# Summary
Ran into some type issues using conditional selections. This PR:
- Updates the conditional selection type inside `grab`
- adds a test case to the `builder` spec

## Technical Details
```ts
import { q, InferType } from 'groqd';

const pokemonQuery = q("*")
  .filter("_type == 'pokemon'")
  .grab(
    {
      _id: z.string(),
    },
    {
      "name == 'Charmander'": {
        name: q.literal("Charmander"),
      },
      "name == 'Bulbasaur'": {
        name: q.literal("Bulbasaur"),
      },
    }
  );
  
// Before

type Pokemon = InferType<typeof query>
//^? { _id: string } | { _id: string; name: "Charmander" } | { _id: string; name: "Bulbasaur"}

type PokemonName = Pokemon[number]["name"]
// ts error: Property 'name' does not exist on type { _id: string } | { _id: string; name: "Charmander" } | { _id: string; name: "Bulbasaur"}


// After

type Pokemon = InferType<typeof query>
//^? { _id: string; name: "Charmander" } | { _id: string; name: "Bulbasaur"}

type PokemonName = Pokemon[number]["name"]
//^? "Charmander" | "Bulbasaur"
```

### DIFF
--- a/src/builder.test.ts
+++ b/src/builder.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { q } from "./index";
+import { q, InferType } from "./index";
 import { z } from "zod";
 import { runPokemonQuery, runUserQuery } from "../test-utils/runQuery";
 import invariant from "tiny-invariant";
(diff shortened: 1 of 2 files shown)


src/builder.test.ts
```diff
@@ -117,11 +117,14 @@ describe("ArrayResult.grab/UnknownResult.grab/EntityResult.grab", () => {
     expect(data[1]).toEqual({ _id: "pokemon.2" });
     expect(data[3]).toEqual({ _id: "pokemon.4", name: "Charmander", hp: 39 });
 
+    type pokemonName = InferType<typeof pokemonQuery>[number]["name"];
+
     for (const dat of data) {
       if ("name" in dat && dat.name === "Charmander") {
-        expect(dat.name === "Charmander").toBeTruthy();
+        const name: pokemonName = dat.name;
```
fritz-c, line 124: Would this change have caused it to fail with the previously implemented type?

littlemilkstudio: yeaup. previously `"name"` wouldn't be indexible on the type `Pokemon` because the base pokemon shape (`{ _id: string }`) would be part of the union.

At the same time tho, @gksander brought up an interesting point in slack.. if it's conditional selection, there's a chance no conditions are met and you're left with and entity that's the shape of the base properties

littlemilkstudio:
Closing this since I'm starting to see how this deviates the type from the groq query

littlemilkstudio:
Closing this since I'm starting to see how this deviates the type from the groq query

PR #80, opened 2023-03-01
Tests for types

gksander:
This PR just decorates/cleans up our test suite a bit with tests regarding expected types. This is long overdue, since one of the main value props of groqd is the type inference you get with it.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,6 +11,7 @@
     "browser": true
   },
   "rules": {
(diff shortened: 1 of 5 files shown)


carlos-kelly (approved):
`toEqualTypeOf` and `expectTypeOf` are neat!

littlemilkstudio (commented):
v nice assertions 👌

PR #82, opened 2023-03-01
Top Level Select Method

littlemilkstudio:
# Summary
Adds `select()` method to several entities:
- `q.select(conditions: { [condition]: Select | Query | [string, ZodType] }): BaseQuery`
- `EntityQuery.select(conditions | select): EntityQuery`
- `ArrayQuery.select(conditions | select): ArrayQuery`

Pulls out some shared utils between `select` & `grab`

I think this establishes some long desired union type handling around empty objects and/or null being in unions returned from a select. **These changes are only implemented in `select()` & do not effect the current select & union types within `grab()`**

## GROQ Scenarios to Consider
```ts
// #1
*[]{
  ...select({ baseProperty })
}

// #2
*[]{
  "nested": nested{
     ...select({ nestedProperty })
   }
}

// #3
*[]{
  "nested": select({ baseProperty })
}
```

This PR accomplishes the following by:
```ts
// #1
const query1 = q("*").filter().select({
  default: {
    baseProperty: q.string()
  }
})

// #2
const query2 = q("*").filter().grab({
  nested: q('nested').select({
    default: {
      nestedProperty: q.string()
    }
  })
})

// #3
const query3 = q("*").filter().grab({
  nested: q.select({
    default: {
      baseProperty: q.string()
    }
  })
})
```

## API considerations
**Standalone Entities**
A big concept I wanted to tackle here was standalone entities to allow breaking up large queries. Currently this is implemented for projections by following this pattern:
```ts
Import { TypeFromSelection, Selection } from 'groqd'

export const pokemon = {
  id: q.string() 
} satisfies Selection

export type Pokemon = TypeFromSelection<typeof pokemon>

// used in another file in a query
const query = q('*')
  .filter('[ _type == "pokemon" ]')
  .grab(pokemon)
```

**Consolidating Type Inference API**
Since `q.__experimental_select()` returns a `BaseQuery`, the `InferType` util will work on it. ie:
```ts
const select = q.select({
  "foo > 5": {
     bar: q.string()
   }
})

type SelectType = InferType<typeof select>
// ^? { bar: string } | null
// null is added here because there is no default condition
```
I figure a nice step forward is allowing all types to be inferred by `InferType<T>` vs having differing generics to derive the types of different things

**Naming**
I tried to keep parity with `groq` as much as possible, so that's why I called it `.select()` which maps to groqs `select()` function. I think this is a little confusing with the existing `Selection` type which is used to represent the type of the argument we pass into the `.grab()` method to create a projection, however, I think it could be nice to keep our terms as close to `groq` long term as possible

## Next Steps
- update `grab(selection, conditions)` to use `select` under the hood for it's conditions
- Add documentation for `select()`
- ~move from `__experimental_select()` -> `select()`~
- Add standalone `grab` support to compose queries replacing `Selection` usage ie:
```ts
const fooProjection = q.grab({
  foo: q.string()
})

type Foo = InferType<typeof fooProjection>

const query = q('*').filter().grab(fooProjection)
```

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -15,8 +15,6 @@ import {
 } from "./nullToUndefined";
 import { Payload, BaseQuery } from "./baseQuery";
 
-export { Payload, BaseQuery } from "./baseQuery";
-
 /**
  * Single Entity
  */
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -1,6 +1,13 @@
 import { z } from "zod";
 import { grab } from "./grab";
+import { select } from "./select";
+import type {
+  ConditionRecord,
+  ConditionSchema,
+  SelectSchemaType,
+} from "./select";
 import type { Selection } from "./types";
+import { extendsBaseQuery } from "./typeGuards";
 import {
   nullToUndefined,
   nullToUndefinedOnConditionalSelection,
@@ -44,6 +51,21 @@ export class NullableBaseQuery<T extends z.ZodTypeAny> extends BaseQuery<
 export class EntityQuery<T extends z.ZodTypeAny> extends BaseQuery<T> {
   constructor(payload: Payload<T>) {
     super(payload);
+  }
+
+  __select<Conditions extends ConditionRecord>(
+    s: Conditions
+  ): EntityQuery<SelectSchemaType<Conditions>>;
+  __select<S extends z.ZodType>(s: BaseQuery<S>): EntityQuery<S>;
+  __select<Conditions extends ConditionRecord>(
+    s: Conditions
+  ): EntityQuery<SelectSchemaType<Conditions>> {
+    const _select = extendsBaseQuery(s) ? s : select(s);
+
+    return new EntityQuery<ConditionSchema<Conditions[string]>>({
+      query: this.query + `{...${_select.query}}`,
+      schema: _select.schema,
+    });
   }
 
   grab<
@@ -118,6 +140,21 @@ export class ArrayQuery<T extends z.ZodTypeAny> extends BaseQuery<
 > {
   constructor(payload: Payload<z.ZodArray<T>>) {
     super(payload);
+  }
+
+  __select<Conditions extends ConditionRecord>(
+    s: Conditions
+  ): ArrayQuery<SelectSchemaType<Conditions>>;
+  __select<S extends z.ZodType>(s: BaseQuery<S>): ArrayQuery<S>;
+  __select<Conditions extends ConditionRecord>(
+    s: Conditions
+  ): ArrayQuery<SelectSchemaType<Conditions>> {
+    const _select = extendsBaseQuery(s) ? s : select(s);
+
+    return new ArrayQuery<ConditionSchema<Conditions[string]>>({
+      query: this.query + `{...${_select.query}}`,
+      schema: _select.schema,
+    });
   }
 
   filter(filterValue = "") {
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -54,13 +54,11 @@ export class EntityQuery<T extends z.ZodTypeAny> extends BaseQuery<T> {
     super(payload);
   }
 
-  __experimental_select<Conditions extends ConditionRecord>(
+  select<Conditions extends ConditionRecord>(
     s: Conditions
   ): EntityQuery<Spread<SelectSchemaType<Conditions>>>;
-  __experimental_select<S extends ZodUnionAny>(
-    s: BaseQuery<S>
-  ): EntityQuery<Spread<S>>;
-  __experimental_select<Conditions extends ConditionRecord>(
+  select<S extends ZodUnionAny>(s: BaseQuery<S>): EntityQuery<Spread<S>>;
+  select<Conditions extends ConditionRecord>(
     s: Conditions
   ): EntityQuery<Spread<SelectSchemaType<Conditions>>> {
     const _select = extendsBaseQuery(s) ? s : select(s);
@@ -145,13 +143,11 @@ export class ArrayQuery<T extends z.ZodTypeAny> extends BaseQuery<
     super(payload);
   }
 
-  __experimental_select<Conditions extends ConditionRecord>(
+  select<Conditions extends ConditionRecord>(
     s: Conditions
   ): ArrayQuery<Spread<SelectSchemaType<Conditions>>>;
-  __experimental_select<S extends ZodUnionAny>(
-    s: BaseQuery<S>
-  ): ArrayQuery<Spread<S>>;
-  __experimental_select<Conditions extends ConditionRecord>(
+  select<S extends ZodUnionAny>(s: BaseQuery<S>): ArrayQuery<Spread<S>>;
+  select<Conditions extends ConditionRecord>(
     s: Conditions
   ): ArrayQuery<Spread<SelectSchemaType<Conditions>>> {
     const _select = extendsBaseQuery(s) ? s : select(s);
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -1,6 +1,6 @@
 import { z } from "zod";
 import { grab } from "./grab";
-import { select } from "./select";
+import { select, spreadUnionSchema } from "./select";
 import type {
   ConditionRecord,
   ZodUnionAny,
@@ -36,7 +36,7 @@ export class EntityQuery<T extends z.ZodTypeAny> extends BaseQuery<T> {
 
     return new EntityQuery<Spread<SelectSchemaType<Conditions>>>({
       query: this.query + `{...${_select.query}}`,
-      schema: _select.schema,
+      schema: spreadUnionSchema(_select.schema),
     });
   }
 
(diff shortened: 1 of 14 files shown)


src/select.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      "foo > 2": {
+        bar: z.boolean(),
+      },
+      default: {
+        baz: z.string(),
+      },
+    });
+
+    expect(query).toBe("select(foo > 2 => { bar }, { baz })");
+    expectTypeOf({} as z.infer<typeof schema>).toEqualTypeOf<
```
fritz-c, line 17: Ooh, cool utility.

src/select.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      default: {
+        baz: z.string(),
+      },
+    });
+
+    expect(query).toBe(
+      'select(foo > 2 => select(bar == "thing" => { b }, { c }), { baz })'
+    );
+    expectTypeOf({} as z.infer<typeof schema>).toEqualTypeOf<
+      { b: string } | { c: string } | { baz: string }
```
fritz-c, line 86: Really impressive how you have all these types resolve.

src/select.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    it("updates query and handles nested object types", () => {
+      const standaloneSelect = q.__experimental_select({
+        "foo > 2": {
+          bar: z.boolean(),
+        },
+        default: {
+          baz: z.string(),
+        },
+      });
+      const entityQuery = q("foo").__experimental_select(standaloneSelect);
```
fritz-c, line 102: ~~So when you call the select method with an instance of the select method's return type, it does the spread behavior? If that's the case, it seems like a pretty obscure way to trigger that behavior (= hard to read the code, hard to write the code).~~
~~Why not have a separate method called `selectAndSpread`/`spreadSelect` or something like that, so you could do the following:~~
```js
q('foo').spreadSelect({
  "foo > 2": {
    bar: z.boolean(),
  },
  default: {
    baz: z.string(),
  },
})
```

src/select.test.ts
```diff
@@ -0,0 +1,222 @@
...
+      const standaloneSelect = q.__experimental_select({
+        "foo > 2": {
+          bar: z.boolean(),
+        },
+      });
+      const entityQuery = q("foo").__experimental_select(standaloneSelect);
+
+      expect(entityQuery.query).toBe("foo{...select(foo > 2 => { bar })}");
+      expectTypeOf({} as z.infer<typeof entityQuery.schema>).toEqualTypeOf<
+        { bar: boolean } | {}
```
fritz-c, line 122: I recall that eslint doesn't like the `{}` type, as [it pretty much means `any`](https://www.typescriptlang.org/play?#code/C4TwDgpgBAQgTgQwMYQM5QLxQN4F8DcAsAFAlID2AdqsFAgFzzJqZQCMZVNUARo4inRZsDAORtRuTtVpJ+zIVFGjp3ACbzBrPKtoRNLLJQCuAG1MkSoSFACiAWzCgA8jwBWEJLSwAlT+Tg1AB4aOABLSgBzABooSggANwg4AD4iUmIKGSgAL3oHJxBXDy9WDkyuWhB8xxd3T28cBmNKNQgAMwiINSkK7IAPGsLihtZlXSgAdyG6ksadPu4EmaL60qMzC2IgA). Any way to work around using it?

src/select.test.ts
```diff
@@ -0,0 +1,222 @@
...
+          baz: z.string(),
+        },
+      });
+      const entityQuery = q("foo").__experimental_select(standaloneSelect);
+
+      expect(entityQuery.query).toBe(
+        "foo{...select(foo > 2 => select(foo == 3 => a, { b }), { baz })}"
+      );
+      expectTypeOf({} as z.infer<typeof entityQuery.schema>).toEqualTypeOf<
+        {} | { b: boolean } | { baz: string }
```
fritz-c, line 161: Since both selects have a default, would the empty object be a possibility here? Also, I don't see the top-level `string` type for `a` accounted for.

src/select.test.ts
fritz-c: I think I see now, you were just demonstrating that the `.select()`'s return type works just fine when you plug them back into a bigger query via another `.select()`. This API is allowed for reusability (yeah, I read your PR notes, but didn't quite grok it at the time). Whether to spread or not is sensitive to the context in which the select is called.

src/select.test.ts
```diff
@@ -0,0 +1,222 @@
...
+          baz: z.string(),
+        },
+      });
+
+    expect(query.query).toBe("*[]{...select(foo > 2 => { bar }, { baz })}");
+    expectTypeOf({} as z.infer<typeof query.schema>).toEqualTypeOf<
+      ({ bar: boolean } | { baz: string })[]
+    >();
+  });
+});
```
fritz-c, line 222: Can you add a test case or two for the non-spreading projection scenario (`#3` in your PR notes)?

fritz-c (commented):
Just partway through this, yet, but since I'm taking forever, I thought I'd pass along the couple of notes I had so far.

fritz-c (commented):
Since the library's still v0.x.x, I think experimental stuff is fine to add without any sort of extra prefixing. It looks pretty solid in the test cases apart from the couple of type questions I had.

gksander:
This is looking really awesome! Still noodling on a couple things, but a couple of things up front:

- For `select.test.ts` can we use one of the test helpers like `runPokemonQuery` and write some assertions against the result of the query (similar to the other tests), to make sure that – not only do the types look like we expect – but that we're also getting data as we'd expect?
- Can you run `yarn changeset` and put a changeset on this branch so we can release this thing once we merge?

src/select.test.ts
littlemilkstudio: yeaup. exactly. Ideally groq would allow something like this:
```
*[]select(...)
```
but it only allows `select` inside a block context, so we get something similar by doing this instead:
```
*[]{
  ...select(...)
}  
```


In the current api prior to this PR, you would accomplish this through the second argument of the `grab` call ie:
```ts
const baseSelection = {
  _type: q.literal('someEntity')
}
const aSelection = {
  _id: q.literal('a')
  a: q.boolean()
}
const conditionB = {
  _id: q.literal('b')
  b: q.number()
}
const selectionTuple = [baseSelection, {
  '_id == "a"': aSelection
  '_id == "b"': bSelection
}]

type SomeEntity = TypeFromSelection<typeof baseSelection> 
| (TypeFromSelection<typeof baseSelection> & (
  | TypeFromSelection<typeof aSelection>
  | TypeFromSelection<typeof bSelection>
))

const query = groq('*')
  .filter('_type == "someEntity"')
  .grab(...selectionTuple)
```

This has its pitfalls tho:
- handling of standalone entities was a little sloppy (need to se a tuple that you spread into the `grab`)
- no support for default condition
- mixed support for empty base selection
- tough to infer the type from the `selectionTuple`
- doesn't handle types well where there's a default condition and empty base (ie sanity field is only ever an `a` entity or a `b` entity)

src/select.test.ts
littlemilkstudio: yea so the way groqd handles spreading on any primitive is that it just returns an empty object, so the empty object is a result of spreading on the ["a", q.string()] field and not a result of the lack of a default...

I think what I've sussed out internally that's going on in groq is that it follows these rules:

no default means null is a possible value if no conditions are met
spreading on a primitive results in an empty object

combine these two together and that's how the empty object shows up when you spread a select and there's no default condition. This assertion is basically asserting the first assumption there in the absence of the second assumption

src/select.test.ts
littlemilkstudio: good catch. yup, will do

src/select.test.ts
fritz-c: Funky behavior on groq's end. Wonder why they don't just make it an error to spread a primitive.
Anyway, thanks for the explanation.

src/select.test.ts
littlemilkstudio: Ended up running into some issues here when testing validation, but updated and wrote some extra tests to cover the fix. We actually did need a `z.record` type in the schema to enforce empty objects in validation:
```ts
const emptyObject = z.record(z.string(), z.never())

// this resolves to record<string, never> like mentioned earlier
type EmptyObject = z.infer<typeof emptyObject>
// ^? Record<string, never>
```

src/select.test.ts
littlemilkstudio: added in the latest commits

fritz-c (approved):
Looks good. Lots of tricky problems solved here.

src/builder.ts
```diff
@@ -1,42 +1,21 @@
...
-}
-
-export class NullableBaseQuery<T extends z.ZodTypeAny> extends BaseQuery<
-  z.ZodNullable<T>
-> {
-  constructor({ schema, query }: Payload<T>) {
-    super({ schema: schema.nullable(), query });
-  }
-}
+export { Payload, BaseQuery } from "./baseQuery";
```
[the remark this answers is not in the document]
littlemilkstudio: ah yea. no they don't. good call

carlos-kelly (approved):
This is great work. Impressive stuff with the types around the Zod Union spread.

Issue #83, opened 2023-03-02
`_type` is not returned with Image Metadata


b3nk3:
Hi there, I've been playing around with groqd today (Amazing - 5⭐) and immediately ran into a problem.

You have the following in [`sanityImage.ts`](https://github.com/FormidableLabs/groqd/blob/main/src/sanityImage.ts#L62)
> `_type: schemas.literal("sanity.imageDimensions"),`

After a few tries I realised, that this field is not returned. (despite some docs on sanity.io)

On their reference however, it is corresponding to how my recent install behaves. [See here](https://www.sanity.io/docs/image-type#df3f768ce073)

gksander:
Thanks for reporting! Admittedly, I just followed their docs when adding the types here. But this should be a pretty quick fix to patch this up! I'll see if I can get something in tomorrow.

gksander:
Was just digging into this, and I'm having troubles replicating the issue – I seem to consistently get the `_type` field back on the image dimensions. Can you give me some more info on the type of image you're querying for that has that `_type` field missing?

FWIW I ran this query in a vision dashboard to check for images with missing `_type` on dimensions:

```
*[_type == "sanity.imageAsset" && defined(metadata.dimensions._type)]
```

and didn't get any results.

b3nk3:
Hiya, sorry, busy day at work here.

Doesn't no results mean it cannot find an image asset where `_type` is defined?

I ran the same query on my Vision - on their movie sample setup and indeed, I get no results. but as soon as i removed `._type` from the query I received the results.

I believe, if you want to find assets, where there is no `_type` field defined, you need to negate the query

```
*[_type == "sanity.imageAsset" && !defined(metadata.dimensions._type)]
```

So basically, you have validated it 😀 by not getting any results.

--------

I still think we perhaps should not remove it from schema creation, but mark it optional, so if there is a weird internal Sanity thing where it may or may not be included, would still allow this to work?

b3nk3:
Okay, I also did a bit more digging - I created my setup from a dataset provided by Sanity... aaand I think when that got loaded it lacked the `_type` field.

I've gone ahead and uploaded an image manually and that has the `_type`.

Despite this, I still think it could be optional. I've created a PR for this. What you do think? #86

PR #86, opened 2023-03-02
feat(sanityImage): make `_type` optional

b3nk3:
Hello 🏄‍♂️!

I made this change on the back of a discussion on  #83.

Basically when working with a dataset from Sanity, I ran into issues with a missing `_type` field on the Image>Dimension property.

This made me think that others might run into it and that this field is not essential, so could be considered optional.

### DIFF
--- a/.changeset/good-fishes-provide.md
+++ b/.changeset/good-fishes-provide.md
@@ -1,0 +1,5 @@
+---
+"groqd": patch
+---
+
+Make `sanityImage` `imageDimensions._type` an optional field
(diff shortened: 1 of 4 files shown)


gksander (approved):
Sweet, this looks great! So that we can release this as a patch, can you run a `yarn changeset` and add in a patch changeset with an appropriate note (e.g. `Make sanityImage imageDimensions._type an optional field`) and then push that up, and then we can get this patch released today!

b3nk3:
Done!

Issue #61, opened 2023-02-10
Expand contentBlock to allow custom fields in Markdefs 


heggemsnes:
We often have an internal link field (reference) in the markDefs part of block content, as of now I can see no way to expand this from the contentBlock helper. 

Would be great to add and option to add additional fields like this: 

```typescript

body: q.array(q.contentBlock({
  markDefs: {
  link: q('internalLink')
    .deref()
    .grab({
      _id: q.string(),
      _type: q.string(),
      slug: ['slug.current', q.string()],
    })
    .nullable(),
  }
}))

```

gksander:
You're absolutely right, this should be allowed! We should be able to get this added in without introducing any sort of breaking changes.

gksander:
Closing this issue, as I believe this has been resolved. If not, we can reopen.

schwartzmj:
> This doesn't seem to work @gksander, in the docs there is an example:
> 
> `markDefs: q.object({ _type: q.literal("link"), href: q.string() })`
> 
> But I'm not sure how that translates to what @heggemsnes had mentioned in the original issue. It's unclear also how this would work for multiple fields to check for different types of links for example (internal, external).
> 
> I get "Invalid input" when trying to pass anything to markDefs, and I also don't see any change to the generated query :(

Same issue here. Unless I'm misunderstanding something, this doesn't appear to work.

jemjam:
Okay, I was considering a new issue, then I found this one. I believe what folks are looking for (or at least what I've been trying) is a way to deref fields within that object schema.

Here's a pretty straight forward example: there's this [great little sanity guide that suggests a pattern for creating internal and external links](https://www.sanity.io/guides/portable-text-internal-and-external-links), where the internal links are references to other documents. The guide [continues to suggest a query](https://www.sanity.io/guides/portable-text-internal-and-external-links#b0847aaef91a) that includes the slug from the referenced document. (My example pictured below has the `reference` (field) called `pointsTo` instead. 😅 ) They have a short breakdown of how to update your query for this to be useful, the last bullet:

> _type == "internalLink" => { "slug": @.reference->slug } If the object in markDefs is of the type “internalLink,” output a key called “slug,” _follow the reference_, and return the document’s slug.

I couldn't find a way to properly recreate this. Within groqd, the markdefs defined within that contentBody() type helper wouldn't allow me following references. You get type errors about not being able to assign a query type to a z/validator type.

[...]

PR #87, opened 2023-03-06
Content block custom markdefs

gksander:
Addresses #61, extends `q.contentBlock` a bit to allow specifying a custom markdef schema.

### DIFF
--- a/src/contentBlock.ts
+++ b/src/contentBlock.ts
@@ -1,0 +1,50 @@
+import { z } from "zod";
+
+/**
+ * Content block schema for standard content blocks.
+ */
+export function contentBlock(): ReturnType<
+  typeof makeContentBlockQuery<typeof baseMarkdefsType>
+>;
+export function contentBlock<T extends z.ZodType>(args: {
+  markDefs: T;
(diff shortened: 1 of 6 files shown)


src/contentBlock.ts
```diff
@@ -0,0 +1,50 @@
...
+  ReturnType<typeof makeContentBlockQuery<typeof baseMarkdefsType>>
+>;
+export function contentBlocks<T extends z.ZodType>(args: {
+  markDefs: T;
+}): z.ZodArray<ReturnType<typeof makeContentBlockQuery<T>>>;
+export function contentBlocks({ markDefs }: { markDefs?: z.ZodType } = {}) {
+  return z.array(makeContentBlockQuery(markDefs || baseMarkdefsType));
+}
+
+function makeContentBlockQuery<T extends z.ZodType>(markDefs: T) {
```
carlos-kelly, line 26: Do we need to prevent  extending `T` from `z.ZodType` allowing collection types? It's my understanding these would not be nested objects.

carlos-kelly (approved):
Neat! 🚢

PR #92, opened 2023-03-15
Docs setup

gksander:
Working on setting up a real docs site that's a bit easier to read/search than a massive README, and hoping to eventually populate it with more examples/guides/best practices.

This PR just sets up docusaurus and puts content in place. Once merged, I'll follow up with Vercel setup to actually deploy the docs – and formideploy to get routing setup so docs eventually live at `formidable.com/open-source/groqd`.


littlemilkstudio (approved):
rad 🎉

PR #94, opened 2023-03-21
feat: accept "image" type in  contentBlock

### LINKED ISSUE #91: Extend image in block content
I can't quite grasp what you have to do to get metadata (palette for example) in an image field in block content. 

In my query I have a `mainImage` where I can get the palette and do stuff with it. In the body there's also an image. What do I need to do to hav that return the palette as well?

```
	mainImage: q
		.sanityImage('mainImage', {
			additionalFields: {
				alt: q.string(),
			},
			withAsset: ['base', 'dimensions', 'palette'],
		})
		.nullable(),
	body: q
		.array(
			q
				.object({
					_type: q.string(),
					_key: q.string().optional(),
					style: q.string().optional(),
				})
				.passthrough(),
		)
		.nullable(),
```

To test I tried (knowing that asset isn't a string):
```
	body: q
		.contentBlocks({
			markDefs: q.object({
				_type: q.literal('image'),
				asset: q.string(),
			}),
		})
		.nullable(),
```
But just got `Invalid literal value, expected "block"`. I can't quite wrap my head around what I would need to do.

b3nk3:
Hey there 🌊

This PR extends the `makeContentBlockQuery` to allow the use of `image` as a block as per Sanity [Docs](https://www.sanity.io/docs/portable-text-editor-configuration#ec55d49cfe6c)

Resolves #91.

### DIFF
--- a/src/contentBlock.ts
+++ b/src/contentBlock.ts
@@ -24,22 +24,32 @@ export function contentBlocks({ markDefs }: { markDefs?: z.ZodType } = {}) {
 }
 
 function makeContentBlockQuery<T extends z.ZodType>(markDefs: T) {
-  return z.object({
-    _type: z.literal("block"),
-    _key: z.string().optional(),
-    children: z.array(
+  return z
+    .object({
+      _type: z.literal("block"),
+      _key: z.string().optional(),
+      children: z.array(
+        z.object({
+          _key: z.string(),
+          _type: z.string(),
+          text: z.string(),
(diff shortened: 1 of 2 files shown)


b3nk3:
sorry about the oversight on the tests - I'll amend them shortly.

scottrippey:
Just to follow up with this, with the 1.0 release, content blocks are created via `q.fragment`, so this no longer applies.

PR #95, opened 2023-03-22
Long time coming: more legible error messages.

gksander:
Right now, when `makeSafeQueryRunner` fails to parse a response (because of schema validation), it just throws a straight-up Zod error. Zod errors contain a lot of info, but can be really hard to read/grok in the terminal. This PR just adds a bit of sugar on top of `z.ZodError` to give a concise indication of the (first) thing that breaks validation. E.g., can get an error message like this:

```txt
Error parsing `result[12].things[5].name`: Expected string, received number.
```

which should be useful, especially with larger lists of content.

### DIFF
--- a/docs/utility-methods.md
+++ b/docs/utility-methods.md
@@ -26,7 +26,7 @@ const data = await runQuery(
 );
 ```
 
-In Sanity workflows, you might also want to pass e.g. params to your `client.fetch` call. To support this, add additional arguments to your `makeSafeQueryRunner` argument's arguments as below.
+In Sanity workflows, you might also want to pass e.g. params to your `client.fetch` call. To support this add additional arguments to your `makeSafeQueryRunner` query executor's arguments as below.
 
 ```ts
 // ...
--- a/docs/utility-methods.md
+++ b/docs/utility-methods.md
@@ -4,6 +4,53 @@ sidebar_position: 5
 
 # Utility Methods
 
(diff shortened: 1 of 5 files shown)


docs/utility-methods.md
```diff
@@ -4,6 +4,53 @@ sidebar_position: 5
...
+});
+export const runQuery = makeSafeQueryRunner((query) => client.fetch(query));
+
+// 👇 Now you can run queries and `data` is strongly-typed, and runtime-validated.
+const data = await runQuery(
+  q("*").filter("_type == 'pokemon'").grab({ name: q.string() }).slice(0, 150)
+);
+```
+
+In Sanity workflows, you might also want to pass e.g. params to your `client.fetch` call. To support this, add additional arguments to your `makeSafeQueryRunner` argument's arguments as below.
```
carlos-kelly, line 29: This sentence is a bit strange, aren't you always using Groqd in a Sanity workflow? Also I think we can drop the , after To support this since it's one complete thought. Finally argument's arguments is a bit dizzying.

carlos-kelly (approved):
These error messages will be great (speaking as someone who has had to become a detective to sort out why). Just one comment about copy in the docs.

docs/utility-methods.md
gksander: Good callout. This is a pretty old sentence, i'll just rework it.

Issue #97, opened 2023-03-22
`grab` forces return type to object


b3nk3:
I'm not sure if it is intentional and/or if I'm doing this correctly with groqd – open to suggestions – but am running into an issue, where my query returns an array of objects, and when I chain a grab onto it, `schema.parse` starts yelling.

Query (works without the grab):
```
images: q("content[_type == 'gallery'].images[].asset")
        .deref()
        .grab$({
          _id: q.string(),
          lqip: ["metadata.lqip", q.string()],
          aspectRatio: ["metadata.dimensions.aspectRatio", q.number()],
          url: q.string(),
        }),
```

Error:
```
errors: [
    {
      code: 'invalid_type',
      expected: 'object',
      received: 'array',
      path: [Array],
      message: 'Expected object, received array'
    }
  ]
```

gksander:
Ahh, so `q()` returns an `UnknownQuery` which assumes the response is _not_ an array. A `.filter` call will flip things from "single entity" to "array of entities". You could probably restructure this to something like:

```ts
images: q("content")
  .filter("_type == 'gallery'")
  .grabOne("images")
  .filter()
  .grabOne("asset")
  .deref()
  .grab$({ ... })
```

but that feels a bit verbose TBH. Perhaps it'd make sense to just add an optional second param to `q()` to explicitly say, "hey, this will start as an array". I'll play around with this idea. Seems like it should be low lift.

gksander:
Okay, recently released `0.12.0` which has an `isArray: true` option for `q()`, so you should be able to do something like this:

```ts
images: q("content[_type == 'gallery'].images[].asset", { isArray: true })
        .deref()
        .grab$({
          _id: q.string(),
          lqip: ["metadata.lqip", q.string()],
          aspectRatio: ["metadata.dimensions.aspectRatio", q.number()],
          url: q.string(),
        })
```

I'm going to close this issue, but if for some reason that `isArray` option isn't working as expected, ping back here and we can reopen this!

PR #98, opened 2023-03-23
`isArray` option for `q()`

gksander:
Addresses #97 by adding an `isArray` option to `q()` so you can pass a more sophisticated base query that returns an array, without having to do a ton of additional method chaining. Can be useful if you want to "mostly just use groq" and then use `.grab$` with schemas to add in types.


carlos-kelly (approved):
Neat 🎉

PR #100, opened 2023-03-23
formideploy setup

gksander:
Setting up formideploy to deploy docs to `formidable.com/open-source/groqd`. Still need to get TF/AWS access keys setup.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,0 +1,54 @@
+name: Deploy Docs Site
+
+on:
+  push:
+    paths:
+      - 'docs/**'
+      - 'website/**'
+  pull_request:
+    branches:
+      - 'main'
+    paths:
+      - 'docs/**'
+      - 'website/**'
+
+jobs:
+  deploy-website:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16
+          cache: 'yarn'
+
+      - name: AWS CLI version
+        run: "aws --version"
+
+      - name: Install Dependencies
+        run: yarn install --frozen-lockfile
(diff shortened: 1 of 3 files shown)


carloskelly13 (approved):
So we're using formideploy here since we're deploying as a sub route of our dotcom over deploying the docs site on Vercel and pointing a subdomain like grodq.formidable.dev.

gksander:
Yep, that's right – for the time being. Once we migrate to the new `formidable.com` architecture, we can do the "zones" approach and just rewrite into the Vercel-deployed site, but in the interim we'll do a formideploy setup to match the other docs configurations.

ryan-roemer:
I don't totally follow the `grodq.formidable.dev`. As long as we don't mess with base url still building always to have `/open-source/groqd/` in all (dev, staging, prod) cases, it doesn't really matter...

ryan-roemer (approved):
LGTM

carloskelly13:
Some of our docs sites for smaller projects exist as subdomains on formidable.dev and we should probably unify in one direction or another.

ryan-roemer:
> Some of our docs sites for smaller projects exist as subdomains on formidable.dev and we should probably unify in one direction or another.

Yeah, that's fine to do if we want to! I'm only really concerned with the tried and true way we've blown off our toes in the past wherein we have dev and/or staging have a root of `DOMAIN.com` and our production be `DOMAIN.com/open-source/NAME`. As long as we adhere to our rule that subpath structure is identical in dev + staging + prod, I'm a happy camper 😄

gksander:
I'd imagine that from an SEO/analytics perspective, having everything under the root domain is the most friendly. But TBH, I was just doing what AMA/Owl/Zephyr did.

carloskelly13:
Agreed. I suspect the subdomain formidable.dev approach was used because it was easier to stand up for our team and get something deployed.

PR #102, opened 2023-03-24
Add `filterByType` convenience method

carlos-kelly:
Add `filterByType` convenience method to receive a single string argument and apply a GROQ filter by type.

### DIFF
--- a/src/builder.ts
+++ b/src/builder.ts
@@ -97,10 +97,7 @@ export class UnknownQuery extends EntityQuery<z.ZodUnknown> {
   }
 
   filterByType(filterTypeValue: string) {
-    this.query += `[_type == '${filterTypeValue}']`;
-    return new UnknownArrayQuery({
-      ...this.value(),
-    });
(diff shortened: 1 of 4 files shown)


src/builder.ts
```diff
@@ -96,6 +96,13 @@ export class UnknownQuery extends EntityQuery<z.ZodUnknown> {
     });
   }
 
+  filterByType(filterTypeValue: string) {
+    this.query += `[_type == '${filterTypeValue}']`;
+    return new UnknownArrayQuery({
+      ...this.value(),
+    });
+  }
```
gksander, line 104:
```suggestion
  filterByType(filterTypeValue: string) {
    return this.filter(`_type == '${filterTypeValue}'`);
  }
```

Just so we don't end up having to duplicate logic with `.filter`, perhaps we just use `this.filter` in this method.

gksander (commented):
Neat. One suggestion to dedup some logic.

gksander (commented):
Quick one

gksander (approved):
Dope

PR #107, opened 2023-03-30
Setup yarn workspaces

gksander:
So that we can _ship more packages_ from this repo.


package.json
```diff
@@ -1,71 +1,31 @@
...
+    "lint": "eslint packages/**/src --quiet",
+    "test": "yarn workspaces run test",
+    "test:watch": "yarn workspaces run test:watch",
     "typecheck": "tsc --noEmit",
-    "lint": "eslint src --quiet",
     "check:ci": "yarn typecheck && yarn lint && yarn test",
-    "build": "tsup",
-    "preversion": "yarn check:ci",
+    "build": "yarn workspaces run build",
     "prepare": "yarn build",
```
carloskelly13, line 16: `prepare` will build the project on `yarn install` as well. Not sure if this is desired, but I confirmed that behavior locally. Pretty fast using `tsup` so I'm not sure it's harmful.

gksander: I migrated that over from what was there. I'm not 100% sure why I have the prepare script there, I'll see if it's necessary or not. (I think with changesets, it's most likely not needed.)

carloskelly13: When I mirrored `groqd`'s changeset config for `react-live` I noticed it since building a FE library took longer and dropped the task as I didn't see any impact to changesets working.

package.json
```diff
@@ -1,71 +1,31 @@
...
     "changeset": "changeset",
-    "version": "yarn changeset version && yarn install --frozen-lockfile"
+    "version": "yarn changeset version && yarn install"
```
carloskelly13, line 18: Also maybe worth adding command forwarding for the docs website at the top level package.

```
"start:docs": "yarn --cwd website start",
"build:docs": "yarn --cwd website build",
```

I don't think the docs website should be part of the monorepo setup since it isn't a shippable package, but no strong feelings either way on this.

carloskelly13 (approved):
Ran through all the commands and this looks good. Just a couple suggestions around the root level `package.json` tasks.

Issue #109, opened 2023-03-30
q.sanityImage returns as any


JohnGemstone:
Hey!

Seems like since v0.12.0 `q.sanityImage` returns `any`. 

Reproduction below, just hover over the types to see how they evaluate:
https://stackblitz.com/edit/react-ts-z3aiuj?file=App.tsx

```tsx
import { z } from 'zod';
import { q } from 'groqd';

const query = q('*').grab({ image: q.sanityImage('image') });

// from groqd@0.12.0 image type is 'any'
type Foo = z.infer<typeof query.schema>;
```

gksander:
Interesting, I'm seeing this as well. Let me take a peak and see what's going on here.

gksander:
Hmm, I've tracked it down to [this change](https://github.com/FormidableLabs/groqd/pull/98/files#diff-a2a171449d862fe29692ce031981047d7ab755ae7f84c707aef80701b3ea0c80). Not sure yet why that overload is causing this, but I've got a place to investigate.

gksander:
Some weird TypeScript stuff going on here, but TL;DR is to import `sanityImage` directly from `groqd` (like `import { sanityImage } from "groqd";`) instead of using `q.sanityImage`. This should be available in `groqd^0.1.4.0`, and should resolve the type issue you're seeing here. Thank you for reporting this! If this, for some reason, doesn't resolve your issue – ping back here so we can reopen this.

JohnGemstone:
Awesome! Importing looks good, thanks again

PR #117, opened 2023-04-06
Fix `q.sanityImage` types

gksander:
#98 introduced a TS overload for the main `pipe` fn, which makes types more complicated. The `sanityImage` also has pretty messy types, and when combined together – the image types just flop to `any`. This PR extracts that `sanityImage` out into its own standalone import, and puts a deprecation notice on the `q.sanityImage` for the time being. This resolves the type issue, and is probably a better approach long term as we continue to do type acrobatics.

Addresses #109

### DIFF
--- a/.changeset/giant-flowers-exist.md
+++ b/.changeset/giant-flowers-exist.md
@@ -1,0 +1,5 @@
+---
+"groqd": minor
+---
+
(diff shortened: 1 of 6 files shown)


carloskelly13 (commented):
Do we want to do a minor version changeset?

gksander:
lol whoops! Yes, def want to do a changeset here. Will push that up

PR #120, opened 2023-04-06
Expose `makeContentBlockQuery` helper

gksander:
This PR exposes a `makeContentBlockQuery` function to give users access to the underlying utility used for `q.contentBlock` in case they need to construct their own block content logic.

### DIFF
--- a/.changeset/soft-beans-visit.md
+++ b/.changeset/soft-beans-visit.md
@@ -1,0 +1,5 @@
+---
+"groqd": patch
+---
+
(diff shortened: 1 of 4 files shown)


fritz-c (approved):

fritz-c:
I realized one thing after the fact, which doesn't ruin the usefulness of the new export, but could make it better. It would be _nice_ if the `markDefs` argument was optional, and defaults as it does in the wrapping functions now.

PR #121, opened 2023-04-06
Expand default error messages

gksander:
This PR just expands the error messaging a bit, by reporting all errors that Zod catches – not just the first.


packages/groqd/test-utils/runQuery.ts
```diff
@@ -3,8 +3,9 @@ import { evaluate, parse } from "groq-js";
 import { pokemonDataset } from "./pokemon";
 import { BaseQuery } from "../src/baseQuery";
 import { userDataset } from "./users";
+import { GroqdParseError, makeSafeQueryRunner } from "../src";
```
gksander, line 6: Changes here are just to use our external-facing `makeSafeQueryRunner` as part of the test running utils.

carloskelly13 (approved):

PR #122, opened 2023-04-06
`makeContentBlockQuery`: optional arg

gksander:
Little sprinkle sprinkle of convenience.

### DIFF
--- a/.changeset/late-seahorses-help.md
+++ b/.changeset/late-seahorses-help.md
@@ -1,0 +1,5 @@
+---
+"groqd": patch
+---
+
+Make `makeContentBlockQuery` arg optional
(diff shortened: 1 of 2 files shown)


fritz-c:
Max explained to me that I could do the query validation/type grabbing as I wanted with
```js
q.select({'_type == "block"': ["{ ... }", q.contentBlock()] })
```
, so there may not be much of a use case for `makeContentBlockQuery` for me in the first place, and it may be misleading to expose it when it can't actually be used as I initially thought it could be used:
```js
q.select({'_type == "block"': makeContentBlockQuery() })
// ^ this won't work. The children and any other array properties do not get fetched
```

gksander:
I see, I see. So in your opinion, is it best to close this PR and then revert the previous related one so that we're no longer exposing that internal method?

fritz-c:
@gksander Yes, sorry for the sudden 180.

gksander:
hey, no worries at all!

PR #124, opened 2023-04-12
(First draft) Playground editor

gksander:
This PR just sets up foundation for groqd playground editor. The idea with this is:

- Monaco is the superior in-browser editor and can handle TS out of the box.
- Bundling Monaco is a _pain_, so I'm proposing we eat that cost by hosting the editor from this repo.
- Playgrounds can then just consume the editor via iframe and communicate via `postMessage`.

This allows us to ship a sanity plugin for groqd playground without forcing the user into a complicated build setup involving monaco.

image


packages/playground-editor/src/messaging.ts
```diff
@@ -0,0 +1,29 @@
+let targetUrl!: string;
```
carloskelly13, line 1: Do we need to non-null assert this? We're only calling the function which is guaranteed to return `string`.

packages/playground-editor/src/App.tsx
```diff
@@ -0,0 +1,139 @@
...
+    };
+  }, []);
+
+  return (
+    <div ref={containerRef} style={{ width: "100vw", height: "100vh" }}></div>
+  );
+}
+
+// Initial code, will likely change in the future.
+const INIT_VALUE = [
```
carloskelly13, line 105: I think this is fine for now, but I wonder if consuming it from a template file might be better and easier to update?

carloskelly13 (approved):
This is pretty neat. I kicked around the tires a bit to see the integrations and I'm impressed it handles auto complete and imports nicely.

Screenshot 2023-04-12 at 4 10 46 PM
Screenshot 2023-04-12 at 4 10 32 PM

packages/playground-editor/src/App.tsx
gksander: yeah agreed. At some point, we'll also allow passing in initial value via query param – so this is going to change in the near future.

PR #125, opened 2023-04-13
Playground setup

gksander:
This PR sets up the tooling to start building out a Sanity plugin for groqd playground. Nothing super pretty here, but wanted to make sure the build/release process was working so we can hit the ground running with the actual playground.


packages/groqd-playground/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "groqd-playground",
```
gksander, line 2: Thoughts on name before we ship the first release?

carloskelly13: Either `playground` or `explorer`

carloskelly13 (approved):

packages/groqd-playground/src/Playground.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from "react";
```
carloskelly13, line 1: Tiny thing, but can't we configure React to be globally included?

packages/groqd-playground/tsup.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from "tsup";
+
+export default defineConfig({
+	entry: ["src/index.ts"],
```
carloskelly13, line 4: Tiny thing but spacing seems off here.

gksander: oh yeah, you're right – I wonder why ESLint didn't pick this up 🤔 I'll take a peak.

gksander: `eslint packages/**/src --quiet` there it is, only checking against `src` files, whoops.

PR #127, opened 2023-04-13
Migrate to pnpm

gksander:
Migrates codebase to use pnpm instead of yarn, which should make things a bit snappier and easier in terms of dependency management. Disregard the failing Vercel deployment, I'll fix that up after this lands.


docs/utility-methods.md
```diff
@@ -194,30 +194,3 @@ q("*")
 
 // -> { cover: { ..., asset: { extension: string; mimeType: string; ...; metadata: { dimensions: { aspectRatio: number; height: number; width: number; }; }; }; } }[]
 ```
-
-## `makeContentBlockQuery`
-
```
gksander, line 199: This is no longer needed, so I'm removing it here (because the change was causing TS errors, and I'm lazy and figured I'd just rip it out instead of fix it and then rip it out).

gksander:
Interesting, I'll look into this. Weird that it didn't show up on my machine, but I'll wipe some things out and figure out what's going on here.

gksander:
Okay @carloskelly13, all tidied up.

carloskelly13 (approved):

PR #128, opened 2023-04-14
Playground functionality

gksander:
She still ugly, but got some of the foundational stuff in place. Will follow up with:

- configuration options for the sanity plugin
- laying out the dashboard in a more useful/appealing way.

https://user-images.githubusercontent.com/12721310/232087888-099e948f-68d7-4595-9ba9-9178392aee45.mp4


packages/groqd-playground/src/Playground.tsx
```diff
@@ -1,9 +1,105 @@
...
+import { z } from "zod";
+import * as q from "groqd";
+import { BaseQuery } from "groqd/src/baseQuery";
+
+export default function GroqdPlayground() {
+  const [query, setQuery] = React.useState<BaseQuery<any>>(q.q(""));
+  const [response, setResponse] = React.useState("");
+  const client = useClient({ apiVersion: "v2021-10-21" });
+
+  const runQuery = React.useMemo(
```
carloskelly13, line 12: does client change or can you make this a ref?

packages/groqd-playground/package.json
```diff
@@ -51,6 +51,11 @@
     "typescript": "^4.9.5"
```
carloskelly13, line 51: do we need styled components here?

carloskelly13 (approved):

packages/groqd-playground/src/Playground.tsx
gksander: I think `client` _can_ change. Right now, memo vs ref probably doesn't make a huge difference – so I can ref this.

PR #129, opened 2023-04-17
Playground panes

gksander:
This PR pretties up the playground UI a bit, using Sanity UI's components. As a follow-up, I'd like to build out a JSON explorer where you can expand/collapse parts of the JSON response, and if there's an error, be able to highlight an error path/message and have it highlight that portion of the raw response. But that's a fair bit of lift, so starting with this for now.

image


packages/playground-editor/src/useDarkMode.ts
```diff
@@ -0,0 +1,30 @@
+import * as React from "react";
+
+const useMediaQuery = (query: string) => {
```
carloskelly13, line 3: Any use for exporting/extracting this out for other usage down the road?

gksander: My thinking on this: keep things concealed until they need to be revealed. If we have a use-case for this down the road, we can slap an `export` on the front and use it.

PR #130, opened 2023-04-17
Move examples to /examples folder

gksander:
(Ignore branch name, decided to split things up a bit and forgot to rename my branch 🙄) Itty-bitty PR, just moves examples to `/examples` folder to make `packages` a little easier to navigate.


carloskelly13 (approved):
I like this. Making `packages/` the code you ship, not just every possible mini project.

PR #134, opened 2023-04-18
More Groqd Playground features

gksander:
Some more improvements to the playground:

- Loading indicator on fetch
- Cmd + Enter in editor will trigger a fetch in the playground
- twoslash-like type hints on `runQuery` to automatically show return type.

image

### DIFF
--- a/.changeset/five-roses-train.md
+++ b/.changeset/five-roses-train.md
@@ -1,0 +1,5 @@
+---
+"groqd-playground": patch
+---
+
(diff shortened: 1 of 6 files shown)


carloskelly13 (approved):
From a UX standpoint, I wonder if we could format it as the object grows it's harder to see

Screenshot 2023-04-18 at 1 49 19 PM

gksander:
Yeah, the code I borrowed intentionally stripped all newlines to keep it to one line – but I agree that it gets a bit hard to read 🤔 I'll noodle some more on that.

PR #136, opened 2023-04-18
Publish playground editor

gksander:
Instead of hosting playground editor on our own Vercel instance, just publish static build files to npm and consume via `unpkg`. This has many benefits, including:

- we don't maintain infra/cost
- we can easily do semver with the iframe url, since unpkg handles that out of the box.

**NOTE:** I've already published this package and updated permissions/tokens, so no changeset file needed here.

### DIFF
--- a/packages/groqd-playground/src/Playground.tsx
+++ b/packages/groqd-playground/src/Playground.tsx
@@ -410,7 +410,7 @@ export default function GroqdPlayground({ tool }: GroqdPlaygroundProps) {
 const IS_DEV = process.env.MODE === "development";
 const EDITOR_ORIGIN = IS_DEV
   ? "http://localhost:3069"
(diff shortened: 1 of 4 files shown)


packages/groqd-playground/src/Playground.tsx
```diff
@@ -410,7 +410,7 @@ export default function GroqdPlayground({ tool }: GroqdPlaygroundProps) {
 const IS_DEV = process.env.MODE === "development";
 const EDITOR_ORIGIN = IS_DEV
   ? "http://localhost:3069"
-  : "https://groqd-playground-editor.formidable.dev";
+  : "https://unpkg.com/groqd-playground-editor@0.0.2/build/index.html";
```
carloskelly13, line 413: Clever!

carloskelly13 (approved):

PR #137, opened 2023-04-18
Tiny fix, start some playground docs

gksander:
Tiny little ENV fix, and start some docs.


docs/groqd-playground.mdx
```diff
@@ -0,0 +1,100 @@
...
+
+```sh
+pnpm add groqd-playground
+```
+
+</TabItem>
+</Tabs>
+
+:::note
+**The playground tool has the following peer dependencies**, most/all of which you should already have installed in a Sanity project:
```
carloskelly13, line 47: I had to read this a couple times. I think we can say it has these peer does, which are typically already a part of most Sanity Studio implementations: icons, etc...

carloskelly13 (approved):
Love this plus the nice tabs. One copy suggestion. Otherwise 🚀

PR #138, opened 2023-04-19
Playground sharing

gksander:
This PR sets some foundation for "sharing" functionality, where you can share a URL that contains the groqd query code. This PR does a few things:

- Stores "raw" code in localStorage and query string, compressing/decompressing using LZ with `lzstring`.
- On instantiation, pull from query string or localStorage to see if we need to init with specific code.
- Add some Share UI to copy url to clipboard.
- Added in a "reset" button, but haven't wired it up yet. That'll reset the code to some default query.

https://user-images.githubusercontent.com/12721310/233117187-301b99f2-6752-412b-b0dc-9b8a571d4b57.mp4


carloskelly13 (approved):
If you wanted to be fancy, why not integrate something like kbar (https://github.com/timc1/kbar) for the commands?

gksander:
That's a nice idea for future us!

PR #139, opened 2023-04-19
What's that? More playground features?

gksander:
A little hard to keep track of everything as I hammer out features, but this one includes:

- Display of query execution time.
- Persisting editor pane width between reloads.
- Reset functionality, where pressing `Reset` button actually resets the code in the editor.
- Minor tweak to type inlay hints in the editor, so that type hint only shows if there's an empty line above a `runQuery` call. (Without this, the type hint can just jump right into the line above it and muddy that line up.)

### DIFF
--- a/packages/groqd-playground/src/Playground.tsx
+++ b/packages/groqd-playground/src/Playground.tsx
@@ -57,6 +57,13 @@ export default function GroqdPlayground({ tool }: GroqdPlaygroundProps) {
     };
   });
   const iframeRef = React.useRef<HTMLIFrameElement>(null);
+  const editorContainer = React.useRef<HTMLDivElement>(null);
+  const editorInitialWidth = React.useMemo(
+    () =>
+      +(localStorage.getItem(STORAGE_KEYS.EDITOR_WIDTH) || 0) ||
+      EDITOR_INITIAL_WIDTH,
+    []
+  );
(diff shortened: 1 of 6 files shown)


packages/groqd-playground/src/Playground.tsx
```diff
@@ -55,6 +58,14 @@ export default function GroqdPlayground({ tool }: GroqdPlaygroundProps) {
       isFetching: false,
     };
   });
+  const iframeRef = React.useRef<HTMLIFrameElement>(null);
+  const editorContainer = React.useRef<HTMLDivElement>(null);
+  const editorInitialWidth = React.useMemo(
+    () =>
+      +(localStorage.getItem(STORAGE_KEYS.EDITOR_WIDTH) || 0) ||
```
carloskelly13, line 65: Probably low risk, but this will still return `NaN` on the left hand if it can't coerce it with `+`

```
> +('1a' || 0)
NaN
```

carloskelly13 (approved):
Neat, just one idea for edge cases.

packages/groqd-playground/src/Playground.tsx
gksander: Interesting. But, luckily, `NaN || EDITOR_INITIAL_WIDTH === EDITOR_INITIAL_WIDTH`, so I think our default fallback should handle that case as well.

PR #140, opened 2023-04-20
Playground: Init with messages

gksander:
Since `unpkg` doesn't preserve query params, we can't initiative with query params. Opting for a postMessage init approach where:

- editor emits a "READY" event to anyone listening.
- playground picks up that "READY" event and responds with an "INIT" payload, including the parent origin.
- editor picks up that "INIT" payload and initiates the editor with it.
- then it's business as usual, editor using parent origin as target for following messages.

### DIFF
--- a/packages/groqd-playground/src/Playground.tsx
+++ b/packages/groqd-playground/src/Playground.tsx
@@ -153,7 +153,6 @@ export default function GroqdPlayground({ tool }: GroqdPlaygroundProps) {
         const payload = messageSchema.parse(JSON.parse(message.data));
 
         if (payload.event === "READY") {
-          message.source;
           const storedCode =
             new URL(window.location.href).searchParams.get("code") ||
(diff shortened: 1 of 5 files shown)


packages/groqd-playground/src/Playground.tsx
```diff
@@ -152,7 +152,18 @@ export default function GroqdPlayground({ tool }: GroqdPlaygroundProps) {
       try {
         const payload = messageSchema.parse(JSON.parse(message.data));
 
-        if (payload.event === "INPUT") {
+        if (payload.event === "READY") {
+          message.source;
```
carloskelly13, line 156: What is this statement doing?

carloskelly13 (approved):

packages/groqd-playground/src/Playground.tsx
gksander: LOL absolutely nothing, good catch.

PR #144, opened 2023-04-21
Playground: JSON Inspector

gksander:
Initial work for JSON inspector. I'm sure it'll break and we can iterate. Right now, showing successful parsed response, and raw response for failed parsing. Errors are highlighted. Can copy response to clipboard (as JSON). Will follow up with ability to click an error from the top panel and have it scroll into view in the JSON explorer.

image

### DIFF
--- a/packages/groqd-playground/src/components/JSONExplorer.styled.tsx
+++ b/packages/groqd-playground/src/components/JSONExplorer.styled.tsx
@@ -1,0 +1,57 @@
+import styled from "styled-components";
+import { Box } from "@sanity/ui";
+
+export const Root = styled.div`
+  font-family: Menlo, monospace;
+  font-size: 0.9em;
+`;
+
+export const Label = styled.span`
+  color: #9d1fcd;
+  @media (prefers-color-scheme: dark) {
+    color: #d05afc;
+  }
+`;
+
+export const Key = styled.span`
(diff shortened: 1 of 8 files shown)


packages/groqd-playground/src/components/JSONExplorer.styled.tsx
```diff
@@ -0,0 +1,58 @@
...
+  hasError?: boolean;
+}>`
+  padding-left: ${({ depth }) => depth * DEPTH_SC}px;
+  border-radius: 4px;
+  cursor: ${({ pointer }) => (pointer ? "pointer" : "initial")};
+
+  background-color: ${({ hasError }) => (hasError ? "#ffe5ea" : "initial")};
+
+  &:hover {
+    background-color: ${({ hasError }) => (hasError ? undefined : "#e7e7e7")};
```
carloskelly13, line 42: does this disable the style or just output `background-color: ;`

carloskelly13 (approved):
Just two questions, overall 🚀

packages/groqd-playground/src/components/JSONExplorer.styled.tsx
gksander: Just disables the style, as to not override the "error" background color!

PR #154, opened 2023-04-25
Playground: Copy Query

gksander:
Changes default editor code (to show how query params can be used), and add functionality to copy query to clipboard.

### DIFF
--- a/packages/groqd-playground/src/components/Playground.tsx
+++ b/packages/groqd-playground/src/components/Playground.tsx
@@ -26,7 +26,7 @@ import { ShareUrlField } from "./ShareUrlField";
 import { useCopyDataAndNotify } from "../util/copyDataToClipboard";
 import { emitInit, emitReset } from "../util/messaging";
 import { JSONExplorer } from "./JSONExplorer";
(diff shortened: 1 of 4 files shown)


packages/groqd-playground/src/components/Playground.tsx
```diff
@@ -439,7 +442,17 @@ export default function GroqdPlayground({ tool }: GroqdPlaygroundProps) {
               <Stack space={3}>
                 <Box>
                   <Box paddingX={3} marginBottom={1}>
-                    <Label muted>Query</Label>
+                    <Label muted>
+                      Query{"  "}
+                      {query.query && (
+                        <span
```
carloskelly13, line 448: Should we give this a role of a button or use a button?

carloskelly13 (approved):

packages/groqd-playground/src/components/Playground.tsx
gksander: ah yeah, probably. This was quick and dirty, but does deserve some aria context. I'll sneak that in here.

PR #155, opened 2023-04-25
Documentation: q.select

littlemilkstudio:
# Summary

- Adds **.select** section to **Query Building**
- Alters documentation for `Conditional selections with .grab` section in **Query Building**

## Test

Check for typos and glance over concepts. This is just a draft, open to changing, removing or adding anything in here.

### DIFF
--- a/.changeset/shy-dots-press.md
+++ b/.changeset/shy-dots-press.md
@@ -1,0 +1,5 @@
+---
+"groqd": patch
+---
+
+Adds documentation for select method
--- a/.changeset/shy-dots-press.md
+++ b/.changeset/shy-dots-press.md
@@ -1,5 +1,0 @@
----
-"groqd": patch
----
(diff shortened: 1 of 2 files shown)


.changeset/shy-dots-press.md
```diff
@@ -0,0 +1,5 @@
+---
+"groqd": patch
+---
+
+Adds documentation for select method
```
gksander, line 5: Since the docs are published separately and not released via NPM, I don't think we'll need a changeset here (once we merge this to main, the new docs will be live).

gksander (approved):
Dope, thank you!

.changeset/shy-dots-press.md
littlemilkstudio: oh rad. I'll just remove this then

PR #162, opened 2023-04-27
Default schema for `.grabOne`

gksander:
Uses `z.unknown()` as default schema for `.grabOne/.grabOne$` if none provided.

### DIFF
--- a/packages/groqd/src/builder.test.ts
+++ b/packages/groqd/src/builder.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, expectTypeOf, it } from "vitest";
 import { q } from "./index";
-import { z } from "zod";
+import { unknown, z } from "zod";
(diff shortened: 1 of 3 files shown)


packages/groqd/src/builder.test.ts
```diff
@@ -261,6 +261,25 @@ describe("grabOne$", () => {
     expect(data[0]).toBeUndefined();
     expect(data[1]).toBeUndefined();
   });
+
+  it("defaults to unknown() if no second arg provided on EntityQuery", async () => {
+    const { data } = await runPokemonQuery(
+      q("*").filterByType("pokemon").slice(0).grabOne$("name")
```
carloskelly13, line 267: til; you could use `grabOne` like this

carloskelly13 (approved):
Neat!

PR #166, opened 2023-04-27
Raw response on `GroqdParseError`

gksander:
Tack the raw response onto `GroqdParseError` to help with debugging validation failures.


carloskelly13 (approved):
This will be great!

PR #168, opened 2023-04-27
Better error highlighting

gksander:
This PR updates the playground to do a better job of error highlighting and jumping, to handle errors beyond just primitive leaf fields.

image


packages/groqd-playground/src/components/Playground.tsx
```diff
@@ -320,22 +326,27 @@ export default function GroqdPlayground({ tool }: GroqdPlaygroundProps) {
...
-            {errorView}
-          </Box>
-          <Box paddingX={3} marginY={3}>
-            <Label muted>Raw Response {execTimeDisplay}</Label>
-          </Box>
-          <JSONExplorer data={rawResponse} highlightedPaths={errorPaths} />
+          <Split mode="vertical">
+            <Flex direction="column" style={{ maxHeight: 400 }}>
+              <Box marginY={3} paddingX={3}>
+                <Label muted>⚠️ Error</Label>
```
carloskelly13, line 332: Not to be a nit, but isn't the ⚠️ more for warnings? Maybe a 🚫 ?

carloskelly13 (approved):

packages/groqd-playground/src/components/Playground.tsx
gksander: "Warning, there's an error" 😆

Issue #57, opened 2023-02-06
Multiple queries in one builder


schwartzmj:
Thank you for your work on this library! I'm new to Sanity and GROQ, so I apologize if this is already answered.

If I want to perform multiple GROQ queries in one request (e.g. get all posts and all faqs), how would I write this with this library?

Related, is it possible to create re-usable portions of queries? For example, if I want to grab an image asset and include the height and width, etc. I could write something like `.partial(myImagePartial)`. Or if I want to re-use my "All Posts" query for a singular "Post" query?

gksander:
Both great questions! I think both should be addressed in our README FAQ, but I'll answer them here first.

## Multiple queires

Yes you can, although it's not immediately obvious how! You should be able to do something like the following:

```ts
q("").grab({
  posts: q("*").filter("_type == 'post'").grab({ name: q.string() }),
  faqs: q("*").filter("_type == 'faq'").grab({ content: q.string() }),
})
```

By leaving the initial query empty, and then having a couple `grab` fields on that, you can create a root-level object with nested queries.

## Reusing portions of queries

Generally, the beef of groqd queries come from the arguments to `grab` calls. You can extract those parts out into `const`s and re-use those, so maybe something like:

```ts
const postContent = { name: q.string() } as const;

// in posts query
q("*").filter("_type == 'post'").grab(postContent);

// in post query
q("*").filter("_type == 'post' && _id == $id").grab(postContent);
```

I think that should go a pretty long way for extracting out reusable bits – but if that doesn't provide the flexibility you're looking for, let us know!

schwartzmj:
Thank you! This is great. Going to give it all a shot.

gksander:
Closing this, and have added these questions to FAQs [here](https://formidable.com/open-source/groqd/faq/).

PR #169, opened 2023-04-28
FAQ: multiple queries

gksander:
Addressing the "multiple queries" question from #57. (The extracting out queries bit has already been documented.)


carloskelly13 (approved):
I also didn't know this. I wonder if this could be better elevated to the `https://formidable.com/open-source/groqd/query-building/` section or something.

PR #170, opened 2023-05-01
Arcade foundation

gksander:
Getting setup for Groqd Arcade. Still ugly, but this PR sets up:

- webpack tweaks
- a page in docusaurus
- tailwindcss
- monaco editor
- foundation for the page, running queries, editing TS/JSON, etc.

image


website/src/arcade/ArcadeEditor.tsx
```diff
@@ -0,0 +1,209 @@
...
+        typeRoots: ["groqd", "zod"],
+        target: monaco.languages.typescript.ScriptTarget.ES5,
+        strict: true,
+        skipLibCheck: true,
+        esModuleInterop: true,
+      });
+
+      monaco.languages.typescript.typescriptDefaults.setExtraLibs(extraLibs);
+
+      const handleContentChange = debounce(runCode, 500);
```
carloskelly13, line 95: Super minor, but you might want to `.cancel()` this in the cleanup, but the odds of this component dismounting mid schedule in 500ms is rare. But it could create a memory leak.

website/src/arcade/ArcadeSuccessView.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react";
+
+type ArcadeSuccessViewProps = {
+  data: unknown;
```
carloskelly13, line 4: Should we type this stronger to something serializable like `Record<string, string | number | boolean | null>` or is that just being silly?

website/src/pages/arcade.tsx
```diff
@@ -0,0 +1,85 @@
...
+          Run query
+        </button>
+      </ArcadeHeader>
+      <Split className="flex-1">
+        <div style={{ width: "50%" }} className="flex flex-col">
+          <ArcadeEditorTabs activeModel={activeModel} switchModel={setModel} />
+          <BrowserOnly fallback={<ArcadeLoadingIndicator />}>
+            {() => {
+              const ArcadeEditor: ArcadeEditorType =
+                // eslint-disable-next-line @typescript-eslint/no-var-requires
```
carloskelly13, line 59: worth disabling this rule package level?

carloskelly13 (approved):
Neat, just some small questions!

website/src/pages/arcade.tsx
gksander: I'll do that if this comes up again. I _think_ this is the only var-require I'll need to have for the site, so I'll just leave it local for now.

PR #172, opened 2023-05-02
Arcade/examples setup

gksander:
Sets up foundation for arcade "examples"

https://user-images.githubusercontent.com/12721310/235748925-efb7376a-0138-483c-98bd-9857cae0c755.mp4


website/src/arcade/examples.ts
```diff
@@ -0,0 +1,66 @@
...
+  "Deref Related Data": {
+    dataset: "pokemon",
+    code: wrapStandardQuery(`
+      q("*")
+				.filterByType("pokemon")
+				.slice(0, 8)
+				.grab({
+					name: q.string(),
+					types: q("types").filter().deref().grabOne("name", q.string()),
+				})`),
```
carloskelly13, line 48: indents feel strange here?

carloskelly13 (approved):
One tiny formatting thing, otherwise ship!

website/src/arcade/examples.ts
gksander: The `js-beautifier` will handle that and clean it up before it's sent to the editor 😄 I was tired of battling with indents, so decided to just write ugly and fast, and automate the beauty.

PR #176, opened 2023-05-04
Arcade: Share Functionality

gksander:
This PR adds in some share functionalities:

- An actual "share" button
- `Cmd + S` will copy share URL to clipboard
- Tidied some things up in the process, removing `useImperativeHandle` for an event emitter setup because I was tired of React fighting me.
- Ability to customize the dataset and share that. When using a preset dataset, it won't store the whole dataset in query params, just the dataset preset key.

In a follow-up, I'll be adding in some toasts to notify user when e.g. share URL was copied to clipboard.

image


website/src/arcade/ArcadeEditor.tsx
```diff
@@ -15,128 +15,97 @@ import { ARCADE_STORAGE_KEYS } from "@site/src/arcade/consts";
...
+    const handleContentChange = debounce(
+      () => runCode({ editor, dispatch }),
+      500
+    );
+    const didChangeInstance = MODELS.ts.onDidChangeContent(handleContentChange);
+
+    editorRef.current = monaco.editor.create(container, {
+      model: MODELS[activeModel],
+      language: "ts",
+      fontSize: 13,
```
carloskelly13, line 73: in the future might be neat to have a set of options that get saved in local storage like font size and a small collection of different font families.

website/src/arcade/share.ts
```diff
@@ -0,0 +1,10 @@
+export const copyToClipboard = (value: string, cb?: () => void) =>
+  navigator.clipboard
+    .writeText(value)
+    .then(cb)
+    .catch(() => null);
```
carloskelly13, line 5: if for some reason we can't copy this, worth throwing up a toast or something? fine as a follow on or as a todo later

carloskelly13 (approved):
Neat! Just a couple feature suggestions for down the road.

website/src/arcade/eventEmitters.ts
```diff
@@ -0,0 +1,30 @@
+class SimpleEventEmitter<S> {
+  cbs: ((x: S) => void)[] = [];
+
+  subscribe(cb: (x: S) => void) {
+    this.cbs.push(cb);
+
+    return {
+      unsubscribe: () => {
+        const indexToRemove = this.cbs?.indexOf(cb) ?? -1;
+        if (indexToRemove >= 0) this.cbs?.splice(indexToRemove, 1);
```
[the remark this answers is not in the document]
gksander: lol I copy-pasted this from another codebase of mine. `cbs` is instantiated with an empty array, so that should be good to go in the first place – I actually think we can just drop the optional chaining, and be confident `cbs` is always existent.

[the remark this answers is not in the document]
gksander: I'll hit that in a follow up PR.

website/src/arcade/ArcadeEditor.tsx
gksander: oh damn, that'd be really neat! noted! (I'd love to have a slick editor font, like fira code or jetbrains mono or something like that.)

PR #177, opened 2023-05-04
Arcade: toasts

gksander:
Add in some toasts with react-super-hot-toasts.

image


website/tailwind.config.js
```diff
@@ -13,5 +13,6 @@ module.exports = {
       padding: "1rem",
     },
   },
+  important: true,
```
gksander, line 16: Tired of battling with docusaurus and others styling, gonna clobber with `important` here, which is somewhat common in these sort of setups.

carloskelly13 (approved):
🍞

PR #182, opened 2023-05-05
Arcade: More Examples

gksander:
Add in some more examples


website/src/arcade/examples.ts
```diff
@@ -63,4 +63,77 @@ export const EXAMPLES = {
...
+            name: q.literal("Bulbasaur"),
+            hp: ["base.HP", q.number()],
+          },
+          // For Charmander, grab the Attack
+          'name == "Charmander"': {
+            _id: q.string(),
+            name: q.literal("Charmander"),
+            attack: ["base.Attack", q.number()],
+          },
+          // For all other pokemon, cast them into an unsupported selection
```
carloskelly13, line 112: should the reader know what an unsupported selection is here?

carloskelly13 (approved):

website/src/arcade/examples.ts
gksander: idk, I just stole this from our docs 😆 I'll add a little note

PR #184, opened 2023-05-08
feat: accept sanity file asset as an image type

rabah:
If you add a `pdf` file to sanity, for example a brochure that could be downloaded on a click of a link, `groqd` currently throws a zod error of: `'result[0].brochure.asset._type': Invalid literal value, expected "sanity.imageAsset"`.
An example query would be: 
```js
brochure: sanityImage('brochure', {
  withAsset: ['base'],
}),
```
We should be able to add a `pdf` and use the sanity image schema to type the response.

I added a Zod Union of `sanity.fileAsset` & `sanity.imageAsset` which should now allow the type of the file to pass. 
There is potentially a little more work to do here as the file type response is not identical to the Image type response, might need some guidance on how you would want to proceed here.

Open to any feedback and happy to make any requested changes.

### DIFF
--- a/.changeset/large-pets-argue.md
+++ b/.changeset/large-pets-argue.md
@@ -1,0 +1,5 @@
+---
+"groqd": patch
+---
+
+ability for sanityImage schema to handle file assets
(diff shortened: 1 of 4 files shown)


gksander (approved):
Awesome – thank you for this!!

Locally, will you run `pnpm run changeset` and add a patch record with a little description of the change here? Then we can merge this through and get it released asap!

rabah:
> pnpm run changeset

Done, should be there now @gksander 😃

gksander (approved):

PR #192, opened 2023-05-12
Add builder `.select$` methods

gksander:
Adds the `nullToUndefined` behavior of `.grab$` methods to the `.select` query methods.

### DIFF
--- a/packages/groqd/src/builder.ts
+++ b/packages/groqd/src/builder.ts
@@ -46,6 +46,21 @@ export class EntityQuery<T extends z.ZodTypeAny> extends BaseQuery<T> {
     });
   }
 
+  select$<Conditions extends ConditionRecord>(
+    s: Conditions
+  ): EntityQuery<Spread<SelectSchemaType<Conditions>>>;
+  select$<S extends ZodUnionAny>(s: BaseQuery<S>): EntityQuery<Spread<S>>;
+  select$<Conditions extends ConditionRecord>(
+    s: Conditions
+  ): EntityQuery<Spread<SelectSchemaType<Conditions>>> {
+    const _select = extendsBaseQuery(s) ? s : select(s, true);
+
+    return new EntityQuery<Spread<SelectSchemaType<Conditions>>>({
+      query: this.query + `{...${_select.query}}`,
+      schema: spreadUnionSchema(_select.schema),
+    });
+  }
+
   grab<
     S extends Selection,
     CondSelections extends Record<string, Selection> | undefined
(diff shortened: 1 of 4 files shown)


carloskelly13 (approved):
Nice! I think for a major release I wonder if making $ behavior, the default makes sense.

packages/groqd/src/select.ts
```diff
@@ -6,9 +6,11 @@ import {
 } from "./selectionUtils";
 import { extendsBaseQuery, isQuerySchemaTuple } from "./typeGuards";
 import type { FromSelection, Selection } from "./types";
+import { nullToUndefined } from "./nullToUndefined";
 
 export const select = <Conditions extends ConditionRecord>(
-  conditionalSelections: Conditions
+  conditionalSelections: Conditions,
+  shouldCoerceNullValues = false
```
littlemilkstudio, line 13: might want to eventually move this to being an options arg if we wanna support something like a `keyOn` option, but this works fine in the interim since it's pretty much only internal

packages/groqd/src/builder.ts
```diff
@@ -46,6 +46,21 @@ export class EntityQuery<T extends z.ZodTypeAny> extends BaseQuery<T> {
     });
   }
 
+  select$<Conditions extends ConditionRecord>(
+    s: Conditions
+  ): EntityQuery<Spread<SelectSchemaType<Conditions>>>;
+  select$<S extends ZodUnionAny>(s: BaseQuery<S>): EntityQuery<Spread<S>>;
+  select$<Conditions extends ConditionRecord>(
+    s: Conditions
+  ): EntityQuery<Spread<SelectSchemaType<Conditions>>> {
```
littlemilkstudio, line 55: does the return type need to respect the `NullToUndefined<T>` generic at all?

packages/groqd/src/select.ts
littlemilkstudio: something to think about here tho is doing something like this:

```ts
const query = q('*').filter().grab({
  foo: q.select({
    '_type == "Type1"': {
      bar: q.string()
    }
  }, true)
})
```
Even tho `shouldCoerceNullValues` is `true` here, the type from this query is:
```
{ foo: ({ bar: string} | null) }[]
```

littlemilkstudio (commented):
I almost wonder if a field level `grab` & `grab$` works better here.

For instance these two queries are equivalent:
```ts
// with this PR
const query1 = q("*")
        .filterByType("pokemon")
        .slice(0)
        .select$({
          "name == 'Bulbasaur'": {
            name: q.literal("Bulbasaur"),
            foo: q.string().optional().default("bar"),
          },
        })

// this works in the existing api        
const query2 = q("*")
        .filterByType("pokemon")
        .slice(0)
        .select({
          "name == 'Bulbasaur'": q('').grab$({
            name: q.literal("Bulbasaur"),
            foo: q.string().optional().default("bar"),
          }),
        })        
```

Would be nice to be have a more normalized api around this practice `q('').grab({...})`, I've used it before to group properties under a key within a selection

gksander:
@littlemilkstudio that's a good idea, using nested `.grab$` calls! I think I like that better here, going to try that out before pulling the trigger on this PR! Thanks for the tip!

PR #201, opened 2023-06-28
chore: update zod to 3.21.4

heggemsnes:
zod@3.20.2 introduced some issues with type recursions that I think may have caused some wierd `Type instantiation is excessively deep and possibly infinite.` in our projects.

Referenced under "Compiler performance" here: https://github.com/colinhacks/zod/releases/tag/v3.21.0

### DIFF
--- a/.changeset/blue-crews-rush.md
+++ b/.changeset/blue-crews-rush.md
@@ -1,0 +1,5 @@
+---
+"groqd": patch
+---
+
+update zod to 3.21.4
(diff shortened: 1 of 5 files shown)


gksander (approved):
Let's give this a shot!

gksander:
@heggemsnes Will you run `pnpm run changeset` and add a minor-version changeset record (and push to this branch) on this so that we can cut a minor release with this change? Then we can get this released ASAP!

heggemsnes:
Done!

PR #205, opened 2023-07-14
Fix syntax error in select section of query-building.md

BrianWalters:
I saw what looks like a syntax issue in this code example so I went ahead and fixed it in this PR. If I'm misunderstanding something please just throw this away!


gksander (approved):
Great find!

PR #207, opened 2023-07-26
GA config for docs

gksander:
Environment variables added in Vercel dashboard.


carloskelly13 (approved):
Nice, will do this for Spectacle.

PR #208, opened 2023-07-27
Arcade: Resizable Sections

littlemilkstudio:
# Summary
Adds resizable sections to groqd arcade

## Visual Updates
https://github.com/FormidableLabs/groqd/assets/35983235/f731cb54-afbe-410e-82c5-58df3d403f96

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
 importers:
 
   .:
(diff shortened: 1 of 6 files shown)


gksander (approved):
Oh man, this is sick! Lots of super fun code in here, a little `lerp`ing gets me all worked up.

One thought, take it or leave it: does it make sense to have some sort of forced min width on the sections so that a section doesn't collapse down into a line? For example, when fully collapsed:

image

CodePen appears to keep a little bit of width, which is nice so you can still tell there's _something_ there.

image

That could 100% be a "nice to have" follow up, no need to block what you've got here.

littlemilkstudio:
> CodePen appears to keep a little bit of width, which is nice so you can still tell there's something there.

Yea I was looking at this too. So codepen doesn't use a min-width, but they put the title in the resizer when it gets below a certain width.

This is definitely something we can do, but would require some style updates so the collapsed section has some visual ties to the expanded section.

The other thing that's a little different with my math is that for 3 sections, I use 2 dividers. Codepen uses 3 dividers for 3 sections, but the far left divider can't move.

We could definitely update it so every section has a divider vs dividers only being between each section, but I thought with the current style of the windows this played better.

**tldr;** it wouldn't be a huge dev lift, but would require some more design thinking. I think for now tho, the line indicating a collapsed section looks clean enough

littlemilkstudio:
> > CodePen appears to keep a little bit of width, which is nice so you can still tell there's something there.
> 
> Yea I was looking at this too. So codepen doesn't use a min-width, but they put the title in the resizer when it gets below a certain width.
> 
> This is definitely something we can do, but would require some style updates so the collapsed section has some visual ties to the expanded section.
> 
> The other thing that's a little different with my math is that for 3 sections, I use 2 dividers. Codepen uses 3 dividers for 3 sections, but the far left divider can't move.
> 
> We could definitely update it so every section has a divider vs dividers only being between each section, but I thought with the current style of the windows this played better.
> 
> **tldr;** it wouldn't be a huge dev lift, but would require some more design thinking. I think for now tho, the line indicating a collapsed section looks clean enough. Wdyt?

Issue #204, opened 2023-07-14
Bug: Return New Class Instance on all Builder Methods


littlemilkstudio:
Noticed we are mutating the query directly in some builder methods like `slice` & `order`. This can cause some irregular behavior depending on usage. Highlighted [in this arcade example](https://formidable.com/open-source/groqd/arcade/?data=pokemon&code=JYWwDg9gTgLgBAbzlArgOwIooKZQJ5wC%2BcAZlBCHAERgA2AhngObnoAmVA3AFCiSyI4ARyKlylKiwhCOPbgGMIaAM7whOfHAC8wgBRUAVFQCU3OHAB0JYLRi4AQngAqeMNn2QA1thBKT3ANRMDTxdMz0qf3MLFnoAI10EcPMAehS4YBI4PAgUODYIOBgAC2BlODjsEmhsABo4NAgYYHlsIuK2xXR4MvzMklxsNBhkuDSYJvpaAAUIb180AC44AG0AAy7h3QASBHVcPAt9-EJjNfqhCzQUEEqoXWMAXQDzcy8fJWXjvFHo5VoWu4AAz1ABMplerxiUHiiV%2BrzQ9BA2C%2BFlUUGAaCYD1q8NOuNGaTgihAyOGcFy8BKvXoaDYcHQJLJ8HiEAAbm0JoT0spsG1gFAoNgmCgGBjmthlKMJjAprN5p9Vhtcltdt8jiFTudhFcbncHs9zKduMZOEA)

We should update all methods to return a new instance of the class with the updated query vs mutating the internal query directly.

PR #211, opened 2023-08-04
Always return new instance

gksander:
Addresses #204, returns new class instance from each of the builder methods.


carloskelly13 (approved):
Neat. Since we want to avoid mutation and side effects is it worth doing a structuredClone or similarly to `this.value()` when we create the new class?

Issue #213, opened 2023-08-10
Support ordering by orderRank field


yhunko:
According to the [Orderable Document list plugin](https://www.sanity.io/plugins/orderable-document-list) documentation, in order to order by the custom rank, we need to use it like so:

```
*[_type == "category"]|order(orderRank)
```

Which seems to not be supported as of now. Current workaround for now is the following usage:

```
export const getQuickAnswersQuery = q('*')
  .filterByType('quickAnswer')
  .grab(quickAnswerSelection)
  // @ts-ignore
  .order('orderRank');
```

Please correct me if I'm misunderstanding something. And if I'm not, it would be great to add a support.

gksander:
Okay, this should be released in `0.15.9`! Thanks for reporting this!

PR #214, opened 2023-08-11
`.order` input type

gksander:
Addresses #213, loosening up the type on the `.order` method a bit – since you can order by more complex expressions than just `FIELD desc`.


carloskelly13 (approved):
Awesome. I've encountered this myself.

Issue #221, opened 2023-09-04
Portable Text issue


tutods:
Hi
I'm trying to use `PortableText` and GROQD, but I'm facing some issues.
This is my block content query:
```ts

const blockContentQuery = (field = 'body') =>
  q(field)
    .filter()
    .select({
      '_type == "block"': ['{...}', q.contentBlock()],
      '_type == "gallery"': galleryQuery,
      '_type == "imageWithAltAndCaption"': imageWithAltAndCaptionQuery(''),
      '_type == "quote"': ['{...}', quoteQuery],
      '_type == "relatedPosts"': [
        "{..., 'posts': posts[]->{title,'slug': slug.current, headline}}",
        relatedPostsQuery,
      ],
      '_type == "videoUrlWithCover"': videoUrlWithCoverQuery(''),
      default: {
        _key: q.string(),
        _type: ['"unsupported"', q.literal('unsupported')],
        unsupportedType: ['_type', q.string()],
      },
    });
```

But the `'_type == "imageWithAltAndCaption"': imageWithAltAndCaptionQuery('')` it's causing me an error.
> Query:
```ts
const imageWithAltAndCaptionQuery = (name = '') =>
  sanityImage(name, {
    additionalFields: {
      alt: q.string(),
      caption: q.string().nullable(),
    },
    withAsset: ['base', 'lqip', 'blurHash'],
    withHotspot: true,
  });
```
> Error:
<img width="951" alt="image" src="https://github.com/FormidableLabs/groqd/assets/18479474/4c3bd318-1601-4e9a-a61b-cc3de8cb7a0a">

Can anyone please help me?

gksander:
Hi @tutods, do you think you could put together a minimal reproducible example in [TS Playground](https://www.typescriptlang.org/play) for me to check out?

tutods:
Hi @gksander 
I'm having some trouble to put all the code in TS Playground
I have tried to put the important code on this CodeSanbox: https://codesandbox.io/s/groqd-yxdxxh?file=/src/services/blog.ts

Can you please check if you need more code?

tutods:
One note:
- If I remove the lines mentioned on code below:
```tsx
const blockContentQuery = (field = 'body') =>
  q(field)
    .filter()
    .select({
      '_type == "block"': ['{...}', q.contentBlock()],
      '_type == "gallery"': [
        `{..., "images": ${imagesListQuery('images').query}}`,
        galleryQuery.schema,
      ],
[...]

PR #223, opened 2023-09-24
fix(sanityImage): `_key` as undefined instead of nullable

tutods:
On this pull request, I'm changing the `sanityImage` types to use `_key` as `undefined` instead of `null`.

I think with this change, the error I have using the `PortableText` (React) will be fixed.
- Issue: **https://github.com/FormidableLabs/groqd/issues/221**

### Error screenshot:
image

### DIFF
--- a/packages/groqd/src/sanityImage.ts
+++ b/packages/groqd/src/sanityImage.ts
@@ -30,7 +30,7 @@ const hotspotFields = {
 };
 
 const refBase = {
-  _key: schemas.string().nullable(),
+  _key: schemas.string().optional(),
   _type: schemas.string(),
 } as const;
 


packages/groqd/src/sanityImage.ts
```diff
@@ -30,7 +30,7 @@ const hotspotFields = {
 };
 
 const refBase = {
-  _key: schemas.string().nullable(),
+  _key: schemas.string().optional(),
```
gksander, line 33: There's a small problem with this. The `refBase` variable gets used in `toGrab` – and `toGrab` is used inside of a `.grab` projection call. This means that the query generated has something like `{ _key }` in it. If you request a field that's absent, the result is `null` – and not `undefined`. Therefore this change would likely throw an error for missing `_key`s.

tutods: Ok I understand, any idea how to fix the error I'm having for example?

PR #239, opened 2023-11-04
Feature: Groq builder 0.1

scottrippey:
# What is `groq-builder`?

This new package is a potential successor to `groqd`.  It has 2 main features:
1. Schema-aware strongly-typed Groq queries.  This provides a better developer experience when writing GROQ (auto-complete) and validates your queries against your schema (via TypeScript).
2.  Runtime validation is optional.  Zod is no longer bundled, but is extremely easy to utilize.

## Why a new package?
Instead of creating this as a GroqD v2, it seemed appropriate to publish as a new package.  While the syntax is extremely similar, and it ought to be easy to upgrade, there's no code reuse.  And GroqD is hard to say 😝 

## This is an Alpha release
This PR is introducing the package, so this is considered an **alpha** release.  It is not yet feature complete, but will be soon.

## TODO:
- [x] Implement core API methods
- [ ] Add CI tasks for tests, types, lint
- [ ] Enable auto-publishing
- [ ] Add README


packages/groq-builder/README.md
```diff
@@ -0,0 +1,92 @@
...
+   .order('price desc')
+   .slice(0, 10)
+   .projection(q => ({
+     name: true,
+     price: true,
+     slug: q.projection('slug').projection('current'),
+     imageUrls: q.projection('images[]').deref().projection('url')
+   }))
+);
+```
```
carbonrobot, line 35: I wonder if a new helper would make this more clear? I feel like `projection` is odd when its a single field selection that outputs to `slug.current`

Perhaps

```
slug: q.projection('slug').select('current'),
// or
slug: q.projection('slug').field('current'),
```

packages/groq-builder/README.md
```diff
@@ -0,0 +1,92 @@
+# `groq-builder`
+
+A **schema-aware**, strongly-typed GROQ query builder.  
+It enables you to use **auto-completion** and **type-checking** for your GROQ queries. 
+
+## Features
+
+- **Schema-aware** - Use your `sanity.config.ts` for auto-completion and type-checking
```
carbonrobot, line 8: What if you don't have access to the sanity.config file?

packages/groq-builder/README.md
scottrippey: This is a great point, and I actually have a lot of potential solutions...
1. Allow strongly-typed dot-notation strings, so we can do `slug: q.projection('slug.current'),`
2. Allow strongly-typed dot-notation strings without wrapping, like: `slug: 'slug.current',`
3. In GroqD, I really like how `q` can be called like `q('slug.current')`, because it's so short!  However, it lacks semantics - you can put any groq text into that string.  
So I've been thinking, maybe `q.projection(...)` could be aliased as `q.p(...)` since it's so commonly used?  It's the only method that I think we'd ever alias.  Having said that ... if we implemented 1. and/or 2., then the problem is really minimized, so maybe it's better to be more verbose?  

WDYT?

packages/groq-builder/README.md
scottrippey: I'll expand this section and add some documentation around this.  

Long story short, if needed, you can **manually** write the types from your schema.  
I know this sounds tedious, but in my experience, all "external schemas" should be documented, and if I'm implementing a query, I'd be more than happy to document the external schema as I'm developing.  This would make my code more resilient, would allow me to mock externals easily, and would not be too difficult.

The schema type is pretty simple; the only complexity is with `references`.  Here's a minimal example:
```ts
declare const references: unique symbol;
type Product = {
  _type: "product",
  name: string,
  price: number,
  category: { _type: "reference", [references]: "category" },
};
type Category = {
  _type: "category",
  name: string,
}
export type SchemaConfig = {
  documentTypes: Product | Category;
  referenceSymbol: typeof referenced;
};
```

packages/groq-builder/README.md
carbonrobot: I like option 1, but I was sure how difficult that would be so I didn't mention it.

danteissaias:
The way the types are defined in the build output (`declare module "../groq-builder"`) causes typescript not to associate those types with the `groq-builder` module.

packages/groq-builder/README.md
scottrippey: I implemented option 1

scottrippey:
@danteissaias Thanks for pointing that out!  I switched the build to use `tsc` instead of `tsup` and it builds correctly now.

PR #248, opened 2023-12-11
groq-builder: compatibility with GroqD

### LINKED ISSUE #247: Failed to resolve entry for package "groq-builder".
When trying `groq-builder` v0.1.3, I get this in my [astro](https://astro.build/) setup:

```
[commonjs--resolver] Failed to resolve entry for package "groq-builder". The package may have incorrect main/module/exports specified in its package.json.
```

My `package.json`:

```
{
  "name": "personal-website",
  "private": true,
  "version": "0.0.1",
  "scripts": {
    "dev": "astro dev",
    "start": "astro dev",
    "build": "astro check && astro build",
    "preview": "astro preview",
    "astro": "astro"
  },
  "dependencies": {
    "@astrojs/check": "0.3.1",
    "@astrojs/tailwind": "5.0.2",
    "@sanity/astro": "2.1.4",
    "@sanity/client": "6.9.1",
    "@sanity/image-url": "1.0.2",
    "astro": "3.6.4",
    "astro-portabletext": "0.9.6",
    "groq-builder": "0.1.3",
    "react": "18.2.0",
    "react-dom": "18.2.0",
    "sanity": "3.20.2",
    "tailwindcss": "3.3.5",
    "typescript": "5.3.2"
  },
  "devDependencies": {
    "daisyui": "^4.4.17"
  }
}
```

Strange thing is that in my VSCode editor, Typescript reports everything ok. No squigly red lines anywhere.

scottrippey:
# What
Adds more features and compatibility with GroqD v0.x, to ensure "upgrading" is as smooth as possible.

For example:
- Optionally adds simple validation methods, like `q.string()` and `q.number().optional()` to assist in migrating from Zod
- Supports `createGroqBuilder<any>()` for a schema-less builder, which still has strong output types

Other improvements
- Improves type mapping, to preserve nullable and array types
- Deprecates `grab/grabOne` in favor of `project/field`
- Fixes #247


packages/groq-builder/package.json
```diff
@@ -54,5 +61,8 @@
   },
   "publishConfig": {
     "provenance": true
+  },
+  "dependencies": {
+    "groq-builder": "^0.1.2"
```
ringods, line 66: @scottrippey `groq-builder` can't depend on itself, no?

scottrippey: Thanks for catching this!

packages/groq-builder/docs/MIGRATION.md
```diff
@@ -0,0 +1,149 @@
...
+    slug: ["slug.current", q.string().optional()],
+    image: q("image").deref(),
+  });
+```
+
+#### After, with `groq-builder`
+
+```ts
+import { createGroqBuilderWithValidation } from "groq-builder";
+const q = createGroqBuilderWithValidation<any>(); // Using 'any' makes the query schema-unaware
```
carbonrobot, line 30: Is `<any>` optional? Can we use it as `createGroqBuilderWithValidation()`?

packages/groq-builder/docs/MIGRATION.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Additional Improvements
+
+### Migrating from `grab -> project` and `grabOne-> field`
+
+First off, Sanity's documentation uses the word "projection" to refer to grabbing specific fields, so we have renamed the `grab` method to `project` (pronounced pruh-JEKT, if that helps). It also uses the phrase "naked projection" to refer to grabbing a single field, but to keep things terse, we've renamed `grabOne` to `field`.  So we recommend migrating from `grab` to `project`, and from `grabOne` to `field`.
+
+Regarding `grab$` and `grabOne$`, these 2 variants were needed to improve compatibility with Zod's `.optional()` utility. But the `project` and `field` methods work just fine with the built-in validation functions (like `q.string().optional()`).
+
+The `grab`, `grabOne`, `grab$`, and `grabOne$` methods still exist, but have been deprecated, and should be replaced with the `project` and `field` methods.
```
carbonrobot, line 139: I recommend making this the first paragraph in this section.

packages/groq-builder/docs/MIGRATION.md
scottrippey: I was thinking, if you "opt-out" of the strongly-typed  schema, then it ought to be explicit.  
If you've got a codebase with `no-any` rules, then you shouldn't be allowed to create a builder without a schema.

packages/groq-builder/docs/MIGRATION.md
```diff
@@ -0,0 +1,149 @@
...
+Since `q` is bound to our Sanity schema, it knows the types of the product's `name`, `price`, and `slug`, so it outputs a strongly-typed result.  And since we trust our Sanity schema, we can skip the overhead of runtime checks.
+
+
+
+
+## Additional Improvements
+
+### Migrating from `grab -> project` and `grabOne-> field`
+
+First off, Sanity's documentation uses the word "projection" to refer to grabbing specific fields, so we have renamed the `grab` method to `project` (pronounced pruh-JEKT, if that helps). It also uses the phrase "naked projection" to refer to grabbing a single field, but to keep things terse, we've renamed `grabOne` to `field`.  So we recommend migrating from `grab` to `project`, and from `grabOne` to `field`.
```
carbonrobot, line 135: I recommend removing the phrase "First off,", which sounds confrontational

packages/groq-builder/docs/MIGRATION.md
```diff
@@ -0,0 +1,149 @@
...
+
+The `grab`, `grabOne`, `grab$`, and `grabOne$` methods still exist, but have been deprecated, and should be replaced with the `project` and `field` methods. 
+
+### `q.select(...)`
+This is not yet supported by `groq-builder`.  
+
+### Validation methods
+
+Most validation methods, like `q.string()` or `q.number()`, are no longer powered by Zod, but they work just the same.  Use Zod if you want to have extra validation logic, like email validation, ranges etc.
+
```
carbonrobot, line 147: Could we provide an example of how to use the old zod methods here for something like email?

packages/groq-builder/docs/MIGRATION.md
scottrippey: I also was considering exporting a loosely-typed `q` too ... because it would be nice to simplify all these examples, and make it even easier to switch from groqd.  But I decided against it because:

- I wanted people to be explicit about using `any` (as I said above)
- I wanted to encourage people to create their own `q` so that if they eventually replace that `<any>` with `<SchemaConfig>` they'd already have all the imports working correctly
- I didn't want IDEs to be confused where to auto-import, because if they have a strongly-typed `q` then they might accidentally import the loosely-typed `q`.

carbonrobot: I agree with all those points, but I wonder if its going to be painful when developers have that import in 50 files (like a prev project we did), and have to update it everywhere.

scottrippey: I didn't say this in the examples, but I guess I was thinking, you'd put this `export const q = createGroqBuilderWithValidation<any>():` in a file like `q.ts`.
Then you change all your `import { q } from 'groqd';` to `import { q } from './q';` instead.  

Should I make this clearer in the docs? Or do you think that's a pattern that folks should decide for themselves?

packages/groq-builder/docs/MIGRATION.md
```diff
@@ -0,0 +1,149 @@
...
+Not all Groq queries can be strongly-typed. Sometimes you need an escape hatch; a way to write a query, and manually specify the result type.
+The `chain` method does this by accepting any `groq` string, and requiring you to specify the result type.  For example:
+
+```ts
+q.chain<{ itemCount: number }>(`
+  { 
+    "itemCount": count(*[_type === "item"])
+  }
+`);
+```
```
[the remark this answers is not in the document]
scottrippey: I was thinking about using `groq` but it didn't feel right, but `raw` seems like a great fit.  I wanted to keep `chain` internal anyway.

packages/groq-builder/docs/MIGRATION.md
carbonrobot: I think it's fine the way you have it. The new import name is a bit long in the tooth, but I don't have a better idea

packages/groq-builder/docs/MIGRATION.md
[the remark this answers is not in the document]
scottrippey: Done

carbonrobot (approved):

PR #251, opened 2023-12-22
groq-builder: validation improvements

scottrippey:
# What

Improves the way we handle validation.  
Enables "tree shaking" to remove validation, if unused.
Improves type-checking of validation methods, for better error detection.


packages/groq-builder/src/validation/index.ts
```diff
@@ -1,14 +1,7 @@
...
 
-export const validate = {
-  ...primitiveValidators,
-  ...sanityValidators,
+export const validation = {
+  ...primitiveValidation,
+  ...sanityValidation,
 };
-
-export function createGroqBuilderWithValidation<TRootConfig extends RootConfig>(
```
scottrippey, line 10: Removed this, in favor of the new syntax:
```
const q = createGroqBuilder().include(validation);
```

scottrippey:
Closed, since #252 already contains this branch.

PR #257, opened 2024-01-23
groq-builder: zod compatibility

scottrippey:
# What

Ensures full compatibility with Zod.  
Ditches the "lite validation" that I had implemented.
If Zod is not used, it can still be tree-shaken with this approach, so it's still an "optional" dependency.

Additional changes:
- Type improvements, documentation as per feedback from @bkonkle 
- Added `validationRequired: true` flag, for enforcing runtime validation if desired
- Removed `$` from `q.conditional$` and `q.select$` methods; these are now just `q.conditional` and `q.select`
- Improved all projections, so that "optional" values are correctly mapped as `| null` types


BrianWalters (approved):
It looks good to me, although a lot of this is over my head. Making Zod tree-shakable is important.

scottrippey:
I added like a hundred more commits :) I had a lot of things on the mind to fix and improve!

heggemsnes:
@scottrippey not sure if this is the right format but using Zod for all validation will certainly slow down the editors and possibly noticable runtime validations as queries become complex. 

Added a bit context in #261 but after using Groqd for >10 production projects we are scrambling for an option as developer experience has become intolerable.

scottrippey:
I'd love to discuss!
For our next major release, we've made zod optional, so maybe that will help with your performance issues. However, we've added a lot more TypeScript in order to enable auto completion too.

I'd love to hear more about the kind of performance issues you've experienced. Are these run time issues? Or TypeScript + IDE performance issues?

PR #258, opened 2024-01-23
groq-builder: use `q.infer()` syntax

scottrippey:
# What

Changes syntax for inferred types.  
Previously, you'd bypass validation and simply infer a type by using `true` in a projection.

Instead, this requires you to use `q.infer()`, making it more obvious what it does.


scottrippey:
Closing this PR. I've reverted these changes, in favor of a different approach (a `validationRequired: true` flag)

PR #259, opened 2024-01-25
groq-builder: WIP cleanup

scottrippey:
# What

Based on feedback, this PR seeks to improve the codebase:
- [x] Clean up, rename, refactor complex types
- [x] Add better documentation around them (jsdocs)
- [ ] Add a CONTRIBUTING.md with instructions and an overview

### DIFF
--- a/packages/groq-builder/src/commands/conditional-types.ts
+++ b/packages/groq-builder/src/commands/conditional-types.ts
@@ -3,13 +3,7 @@ import {
   ProjectionMap,
   ProjectionMapOrCallback,
 } from "./projection-types";
-import {
-  Empty,
-  IntersectionOfValues,
-  Simplify,
-  Tagged,
-  ValueOf,
-} from "../types/utils";
+import { Empty, IntersectionOfValues, Simplify, ValueOf } from "../types/utils";
 import { ExtractTypeNames, RootConfig } from "../types/schema-types";
 import { GroqBuilder } from "../groq-builder";
 import { IGroqBuilder, InferResultType } from "../types/public-types";
@@ -25,7 +19,12 @@ export type ConditionalProjectionMap<
       ) => ProjectionMap<TResultItem>);
 };
 
-export type ConditionalExpression<TResultItem> = Tagged<string, TResultItem>;
+/**
+ * For now, none of our "conditions" are strongly-typed,
+ * so we'll just use "string":
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export type ConditionalExpression<TResultItem> = string;
 
 export type ExtractConditionalProjectionResults<
   TResultItem,
--- a/packages/groq-builder/src/commands/conditional.ts
+++ b/packages/groq-builder/src/commands/conditional.ts
@@ -13,7 +13,7 @@ import { ProjectionMap } from "./projection-types";
 declare module "../groq-builder" {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   export interface GroqBuilder<TResult, TRootConfig> {
-    conditional$<
+    conditional<
       TConditionalProjections extends ConditionalProjectionMap<
         ResultItem.Infer<TResult>,
         TRootConfig
@@ -32,7 +32,7 @@ declare module "../groq-builder" {
 }
 
 GroqBuilder.implement({
-  conditional$<
+  conditional<
     TCP extends object,
     TKey extends string,
     TIsExhaustive extends boolean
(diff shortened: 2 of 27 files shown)


packages/groq-builder/src/commands/conditional.ts
```diff
@@ -13,9 +13,9 @@ import { ProjectionMap } from "./projection-types";
 declare module "../groq-builder" {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   export interface GroqBuilder<TResult, TRootConfig> {
-    conditional$<
+    conditional<
```
bkonkle, line 16: Do you know what the `$` signified before, or was it just there out of habit? :sweat_smile:

packages/groq-builder/src/commands/conditional.ts
```diff
@@ -24,15 +24,15 @@ declare module "../groq-builder" {
       conditionalProjections: TConditionalProjections,
       config?: Partial<ConditionalConfig<TKey, TIsExhaustive>>
     ): ExtractConditionalProjectionResults<
-      ResultItem<TResult>,
+      ResultItem.Infer<TResult>,
```
bkonkle, line 27: I like the way this self-documents - it makes it easier to understand what's happening. :+1:

packages/groq-builder/src/commands/filterByType.ts
```diff
@@ -1,13 +1,16 @@
...
   export interface GroqBuilder<TResult, TRootConfig> {
-    filterByType<TType extends ExtractTypeNames<ResultItem<TResult>>>(
+    filterByType<TType extends ExtractTypeNames<ResultItem.Infer<TResult>>>(
       ...type: TType[]
     ): GroqBuilder<
-      ResultOverride<TResult, Extract<ResultItem<TResult>, { _type: TType }>>,
+      ResultItem.Override<
+        TResult,
+        Extract<ResultItem.Infer<TResult>, { _type: TType }>
+      >,
```
bkonkle, line 13: If you want, you could probably import `Override` locally so you don't have to specify the `ResultItem.` prefix. I do this often in Rust, only using the namespace if there are conflicting symbols. So - if there's another `Override` type you're working with you'd use the fully-qualified form (`ResultItem.Override`) for both of the conflicting symbols, but otherwise you could just use `Override`.

I think it's as easy as `import Override = ResultItem.Override`, but you can decide whether it's worth doing or would be confusing.

packages/groq-builder/src/types/result-types.ts
```diff
@@ -1,51 +1,113 @@
...
+ *
+ * However, most chainable methods need to look at the shape
+ * of the inner "ResultItem", ignoring the nullable and Array wrappers.
+ * You can do this with `ResultItem.Infer<TResult>`.
+ *
+ * Additionally, these methods might need to rewrap a new type,
+ * while preserving the nullable/Array wrappers.
+ * You can do this with `ResultItem.Override<TResult, TResultItemNew>`
+ */
+export namespace ResultItem {
```
bkonkle, line 17: I like how this turned out a lot!

packages/groq-builder/src/commands/conditional-types.ts
```diff
@@ -25,7 +19,12 @@ export type ConditionalProjectionMap<
       ) => ProjectionMap<TResultItem>);
 };
 
-export type ConditionalExpression<TResultItem> = Tagged<string, TResultItem>;
+/**
+ * For now, none of our "conditions" are strongly-typed,
+ * so we'll just use "string":
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export type ConditionalExpression<TResultItem> = string;
```
bkonkle, line 27: This makes sense to me. There might be some value in differentiating between different kinds of strings, but only if there's a strong risk of them getting mixed up with each other.

packages/groq-builder/src/commands/conditional.ts
scottrippey: Initially, there were a few methods that had "loose types". I wanted a way to distinguish those methods, so that it was easy to tell that you weren't fully type safe. 
For example,  these conditional expressions are just strings, so they're not safe. 
I used the `$` to mark these unsafe methods. 
However,  now that I've implemented nearly the full api, I've realized that the only "loose types" are these conditions. So ultimately,  I don't think it's worthwhile to use a naming convention to indicate this.  I'd rather use documentation.

packages/groq-builder/src/commands/filterByType.ts
scottrippey: In this case,  I prefer the semantics of "Result Item Override" because that's more descriptive of what's happening.

packages/groq-builder/src/commands/conditional-types.ts
scottrippey: I kinda added this as a placeholder. A conditional expression can be any complex expression,  so there's no way I could enforce it via template strings. But,  I was thinking that I could enable some common auto completions.
For now, this is just a nice way for me to track all the places that were using strings for expressions.

scottrippey:
@bkonkle  Thank you for all the feedback so far!  I've merged all this work into my `zod-again` branch, to consolidate it all.

PR #278, opened 2024-03-14
upgrade vitest from 0.x to 1.x

### LINKED ISSUE #274: groqd: remediate critical and high security warnings
https://github.com/FormidableLabs/groqd/security/dependabot?q=is%3Aopen+severity%3Acritical%2Chigh

Please tag @scottrippey in PR reviews

Burnett2k:
### Description

This upgrades our devDependency on vitest from a beta version to the official 1.x.x major release. This meant snapshots needed to be updated, and we have to modify some of the types a little bit in the PR.

Kudos to @scottrippey for helping me get past a couple of the weird TS issues encountered.

Once this is done, we should no longer be referencing a vulnerable vite server vulnerability AND will be on a more stable version of Vitest for future testing.

After this + updating Sanity to `3.15.0`, we're not pointing at the vulnerable version of `4.2.1` of vite server any longer.

```
❯ pnpm why vite -r
Legend: production dependency, optional only, dev only

playground-example@0.0.1 /Users/sawyerburnett/git-repos/formidable/groqd/examples/playground-example

dependencies:
groqd-playground link:../../packages/groqd-playground
└─┬ sanity 3.15.0
  ├─┬ @vitejs/plugin-react 4.2.1
  │ └── vite 4.5.2 peer
  └── vite 4.5.2
sanity 3.15.0
├─┬ @vitejs/plugin-react 4.2.1
│ └── vite 4.5.2 peer
└── vite 4.5.2

groq-builder@0.9.0 /Users/sawyerburnett/git-repos/formidable/groqd/packages/groq-builder

devDependencies:
vitest 1.3.1
├── vite 5.1.6
└─┬ vite-node 1.3.1
  └── vite 5.1.6

groqd@0.15.10 /Users/sawyerburnett/git-repos/formidable/groqd/packages/groqd

devDependencies:
vitest 1.3.1
├── vite 5.1.6
└─┬ vite-node 1.3.1
  └── vite 5.1.6

groqd-playground@0.0.18 /Users/sawyerburnett/git-repos/formidable/groqd/packages/groqd-playground

dependencies:
sanity 3.15.0
├─┬ @vitejs/plugin-react 4.2.1
│ └── vite 4.5.2 peer
└── vite 4.5.2
```

Fixes #274 

#### Type of Change

- [x] security update

- [x] I have included a changeset if this change will require a version change to one of the packages.
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] Any dependent changes have been merged and published in downstream modules


packages/groq-builder/src/commands/conditional.test.ts
```diff
@@ -68,13 +68,13 @@ describe("conditional", () => {
   it("the query should look correct", () => {
     expect(qAll.query).toMatchInlineSnapshot(
       `
-      "*[_type == \\"variant\\"] {
+      "*[_type == "variant"] {
```
Burnett2k, line 71: Almost all of the snapshot changes are because v1.0 no longer escapes quotes. More info [here](https://vitest.dev/guide/migration.html#snapshots-update-3961)

packages/groq-builder/src/commands/project.test.ts
```diff
@@ -650,8 +650,8 @@ describe("project (object projections)", () => {
 
       await expect(() => executeBuilder(qParser, { datalake: invalidData }))
         .rejects.toThrowErrorMatchingInlineSnapshot(`
-        "1 Parsing Error:
-        result[5].price: Expected number, received string"
+        [ValidationErrors: 1 Parsing Error:
```
Burnett2k, line 653: It appears that the snapshot changed here to include the ErrorType. At this point it is unclear why. Will investigate.

packages/groq-builder/src/commands/project.test.ts
```diff
@@ -669,7 +669,7 @@ describe("project (object projections)", () => {
           price: true,
         })
       ).toThrowErrorMatchingInlineSnapshot(
-        '"[groq-builder] Because \'validationRequired\' is enabled, every field must have validation (like `q.string()`), but the following fields are missing it: \\"price\\""'
+        `[TypeError: [groq-builder] Because 'validationRequired' is enabled, every field must have validation (like \`q.string()\`), but the following fields are missing it: "price"]`
```
Burnett2k, line 672: It appears that the snapshot changed here to include the ErrorType. At this point it is unclear why. Will investigate.

packages/groq-builder/src/commands/select.test.ts
```diff
@@ -186,9 +186,9 @@ describe("select", () => {
     it("should fail with invalid data", async () => {
       await expect(() => executeBuilder(qSelect, invalidData)).rejects
         .toThrowErrorMatchingInlineSnapshot(`
-        "2 Parsing Errors:
+        [ValidationErrors: 2 Parsing Errors:
```
Burnett2k, line 189: It appears that the snapshot changed here to include the ErrorType. At this point it is unclear why. Will investigate.

packages/groq-builder/src/commands/deref.test.ts
```diff
@@ -17,11 +17,12 @@ describe("deref", () => {
   const qVariants = qVariantsRefs.deref();
 
   it("should deref a single item", () => {
-    expectTypeOf<
-      InferResultType<typeof qCategory>
-    >().toEqualTypeOf<SanitySchema.Category | null>();
+    // I don't understand why this is now failing.
+    // switching to toMatchTypeOf seems to fix it, but that does not seem like the correct solution
+    type CategoryInferenceResult = InferResultType<typeof qCategory>;
+    expectTypeOf<CategoryInferenceResult>().toEqualTypeOf<SanitySchema.Category | null>();
```
Burnett2k, line 23: @scottrippey This failure is really throwing me for a loop. I cannot understand why this is failing. I know you're busy on a client project, but if you get time I could use some guidance on what might be happening here. Otherwise, I can try and take another stab at it next week once I've had a bit of a brain break.

Upgrading to vitest caused a few other problems when using `expectTypeOf` in addition to `toEqualTypeOf` and it is unclear to me if this is due to it finding an actual problem with our types, or if they have a bug on their end (vitest's).

I've googled this quite a bit and not coming back with any promising leads.

[Here's a link to one of the failures](https://github.com/FormidableLabs/groqd/actions/runs/8302614759/job/22725239327#step:9:19)

It appears if I convert the test to check an array of categories like the variants test below, it all works fine. But something to do with the slice operation and checking a single item seems to cause issues.

scottrippey: Debugging these type errors is hard. Vitest gives a very cryptic error message ("expected 1 argument but got 0").
This error simply means that the types didn't match.
To find the actual issue, you've got to do some debugging in TypeScript, which is a unique experience. I'll take a look locally and see what I can find!

packages/groq-builder/src/commands/conditional.test.ts
scottrippey: Oh this makes me really happy. It's so much easier to read!

packages/groq-builder/src/commands/select.test.ts
scottrippey: That seems fine to me. It's a better snapshot.

packages/groq-builder/src/commands/deref.test.ts
scottrippey: I found the issue, it looks like with this update,  `{ foo: any } & { bar: any }` is no longer the same as `{ foo: any, bar: any }`.
I improved one of my types upstream, and it fixed this problem here.

Burnett2k: @scottrippey Thanks for taking a look. I think we were pushing at the same time, so didn't notice until now but your commit appears to have fixed that issue, but created a new type error in the build step :(.

```
src/tests/mocks/nextjs-sanity-fe-mocks.ts:130:3 - error TS4053: Return type of public method from exported class has or is using name 'decorator' from external module "/Users/sawyerburnett/git-repos/formidable/groqd/packages/groq-builder/src/tests/schemas/nextjs-sanity-fe.generated" but cannot be named.

130   generateSeedData({
      ~~~~~~~~~~~~~~~~

src/tests/mocks/nextjs-sanity-fe-mocks.ts:130:3 - error TS4053: Return type of public method from exported class has or is using name 'referenced' from external module "/Users/sawyerburnett/git-repos/formidable/groqd/packages/groq-builder/src/tests/schemas/nextjs-sanity-fe.generated" but cannot be named.

130   generateSeedData({
      ~~~~~~~~~~~~~~~~


Found 2 errors in the same file, starting at: src/tests/mocks/nextjs-sanity-fe-mocks.ts:130
```

I did go back to detached state with your commit and verified when running `pn run build` on `/groq-builder` we now get that error. I will see if I can find a solution to it. It seems decorator is a reserved word and is now causing an issue 🤷🏻

packages/groqd/vitest.config.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export default defineConfig({
+  test: {
+    typecheck: {
+      enabled: true,
+      checker: "tsc",
+      allowJs: false,
+      include: ["**.test.ts"],
+    },
+    exclude: [...configDefaults.exclude],
```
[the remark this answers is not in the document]
Burnett2k: This means now if there's type errors we should see it fail when running tests locally, rather than on the lint and type check step in CI.

scottrippey (approved):
Awesome work, thank you so much for persevering!

PR #283, opened 2024-04-19
Bump sanity/ui to version 2 in groqd-playground

carbonrobot:
Bumps sanity ui to version 2 and supports multiple versions of styled-components.

There is a version mismatch in sanity packages where @sanity/ui requires styled-components@^6 and @sanity requires styled-components@^5. The features we use are compatible across both versions so we allow both as a peer dependency to resolve warnings when consumers install our playground package.

- Updates sanity/ui peer dependency to v2
- Updates the unpkg playground reference to v0.0.6
- Updates the README file for new branding


.npmrc
```diff
@@ -1,2 +1,3 @@
 strict-peer-dependencies=false
 prefer-workspace-packages=true
+auto-install-peers=true
```
carbonrobot, line 3: This was already set in the lock file, likely due to someone having it set per-machine. Setting it here so it applies to all developers that run `pnpm i`

packages/groqd-playground/package.json
```diff
@@ -50,12 +50,12 @@
...
   "peerDependencies": {
     "@sanity/icons": "^2.3.1",
-    "@sanity/ui": "^1.3.2",
+    "@sanity/ui": "^2",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-is": "^18.2.0",
     "sanity": "^3.15.0",
-    "styled-components": "^5.3.9"
+    "styled-components": "^5.2 || ^6"
```
scottrippey, line 58: ❤️ (fix for issues with the latest Sanity version)

scottrippey (approved):

Issue #323, opened 2025-01-17
Applying `.filter` to a `.deref` value produces incorrect GROQ query


zackdotcomputer:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Bug report

Expectation:
Inside of a projection, when dealing with an array field, it should be possible to dereference and filter that field as follows:
```
// ... Rest of .project call
someField: s.field("someField[]").deref().filter("isPublic")
```

However, this produces the GROQ string:
`"someField": someField[]->[isPublic]`

Groq interprets this as an attempt to get the item at the isPublicth spot in the list and returns null.

To get a valid result, groqd should generate:
`"someField": (someField[]->)[isPublic]`

This is properly interpreted as a filter being applied to the list after dereferencing.

scottrippey:
Good find, thank you! Investigating a fix.

scottrippey:
I've published a new version with a fix.  
The fix includes a little bit of logic to fix this specific scenario (`.deref().filter()`), and adds a unit test.

But it makes me wonder if there are additional scenarios that need to be considered.  Are there other queries that can come after a `deref` that are ambiguous?  Are there other scenarios where `.filter` is ambiguous?  
If you can think of any other examples, please let me know, and I'll create a unit test to reproduce.

zackdotcomputer:
I wonder if the safest way to solve it is just always to wrap the thing being filtered in `(...)` when serializing a `.filter` to groq? So it would always be that `a.filter(b)` would serialize to `(a)[b]`.

I think the issue here really is groq's for having an ambiguity in its grammar but at least the hack of using a parenthetical seems to resolve it?

scottrippey:
Well, with this PR, I added a private `wrap` utility, so maybe I will make that a public api so it's easier to work around future issues.

PR #329, opened 2025-01-30
Fix: support filters after a deref

scottrippey:
# What
When calling `.deref().filter(...)`, this code correctly wraps the deref expression in parenthesis.

Addresses #323.

### DIFF
--- a/packages/groqd/src/types/groq-expressions.test.ts
+++ b/packages/groqd/src/types/groq-expressions.test.ts
@@ -203,14 +203,16 @@ describe("Expressions", () => {
 });
 describe("Expressions.Conditional", () => {
   type T = Expressions.Conditional<FooBarBaz, QueryConfig>;
-  type Expected =
-    | "foo == (string)"
-    | `foo == "${string}"`
-    | `bar == (number)`
-    | `bar == ${number}`
-    | `baz == true`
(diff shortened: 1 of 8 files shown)


packages/groqd/src/types/groq-expressions.test.ts
```diff
@@ -201,3 +201,16 @@ describe("Expressions", () => {
     });
   });
 });
+describe("Expressions.Conditional", () => {
```
sblinde, line 204: I don't know if the reason why the test is failing is because there's no `it(...` line here, so it doesn't realize there's a test here? But that seems to be why that suite is failing, because you have a test written out here.

sblinde:
I'm less fuzzy on Groq but from what I've reviewed, this looks sound. I think there's one specific thing you should check out with the test that might be why the unit test is failing perhaps? Otherwise, if my lack of the familiarity with this repo just means this is a known issue with this repo and that's par for the course, I can give an approval, just let me know.

packages/groqd/src/types/groq-expressions.test.ts
scottrippey: Thank you for catching this! Fixed!

sblinde (approved):
LGTM!

Issue #268, opened 2024-02-17
groq-builder Make slice(0) nullable by default


heggemsnes:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Feature Request

Since we are not really reading validation on "min" for arrays, doing slice(0) should always infer to be nullable. 

Also with document queries, i.e:

`const pageQuery = q.star.filterByType("page").filter("slug.current == $slug").slice(0)`

groq will return null if the document doesn't exist. I feel like this should be the default. 

We can add .nullable() but I think it is more correct to make it default nullable in this scenario and perhaps add a .nonNullable() if necessary.

scottrippey:
This is a great point.   I like the idea of adding a `.notNull()` assertion too!

PR #331, opened 2025-01-31
Feature: `notNull`

scottrippey:
# What
Addresses #268

Makes `.slice(0)` return a nullable result.

Adds a `.notNull()` helper, for asserting that queries do not return null.  Throws an error if `null` is detected.

Additionally, it updates the `.nullable()` utility: if the query is already detected as nullable, then you must be explicit about allowing the redundancy by passing `.nullable(true)` instead.


sblinde (approved):
Solid. I don't see any deal breakers here and really well documented along the way too.🚀

PR #338, opened 2025-02-05
feat(score): adding score command

### LINKED ISSUE #339: Support for `score`
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Feature Request

Hi me again 🙈 

the `score` function from [v0.x](https://commerce.nearform.com/open-source/groqd/docs/legacy/query-building#score) is missing in the v1.x version. It would be very nice if we can implement it.

I can implement it with
```TS
q.star
  .filterByType("pokemon")
  .raw("| score(...)")
```

But that will opt out of typing. The `.order('_score desc')` command does not know the `_score` field and any `.project({...})` will result in a typescript error that I can only comply with `// @ts-ignore`.

I added a PR #338 that might help with the implementation.

matzexcom:
### Description
Adding the groq [score function](https://www.sanity.io/docs/groq-functions#4798b2cba8df) as a command that adds a pipe score function to the query and adds the new field `_score` to the list of keys in the ResultItem.

There may be two significant things to improve or review here.

First, I do not know if extending the TResult with
```TS
Array<ResultItem.Infer<TResult> & { _score: string }>
```
is the right way to do it.

Secondly, regarding the typing (and probably the naming) of the `scoringString` input parameter, I copied the type from the `filter` method and extended it with the match expression.

So we have to give some responsibility back to the consumer.

The order of the chaining of the commands may also have to be considered.

Fixes #339 

#### Type of Change
- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

### How Has This Been Tested?
I added a test file and tested it together with the order command. That might be a violation, but it checks whether the `_score` field is present and correctly filled.

- [ ] I have included a changeset if this change will require a version change to one of the packages.
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/packages/groqd/src/commands/filter.test.ts
+++ b/packages/groqd/src/commands/filter.test.ts
@@ -8,7 +8,7 @@ const qVariants = q.star.filterByType("variant");
 
 describe("filter", () => {
   it("should allow for untyped filter expressions", () => {
-    const qAnything = qVariants.filter("ANYTHING");
+    const qAnything = qVariants.filterRaw("ANYTHING");
     expectTypeOf<InferResultType<typeof qAnything>>().toEqualTypeOf<
       Array<SanitySchema.Variant>
     >();
@@ -18,7 +18,7 @@ describe("filter", () => {
   });
 
   const qFiltered = qVariants
-    .filter('name == "FOO"')
+    .filterRaw('name == "FOO"')
     .project({ name: true, id: true });
   const data = mock.generateSeedData({
     variants: [
@@ -29,7 +29,7 @@ describe("filter", () => {
   });
 
   it("should not change the result type", () => {
-    const qFiltered = qVariants.filter("ANYTHING");
+    const qFiltered = qVariants.filterRaw("ANYTHING");
     expectTypeOf<InferResultType<typeof qVariants>>().toEqualTypeOf<
       InferResultType<typeof qFiltered>
     >();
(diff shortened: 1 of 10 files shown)


matzexcom:
Thanks again for the quick reply :)
Of course, you can take this PR and update it.
It felt a bit quick and dirty to merge the `_score` this way, so I am with you with a more sophisticated approach (besides that, I couldn't 100% follow your explanation)

scottrippey:
Ok so I added a lot of updates.  I think I love hijacking PRs. I'll share what I did here.

First off ... in my explanation above, I was totally wrong about how `_score` worked.  You were 100% correct to add it to the result type via `ResultItem.Infer<TResult> & { _score: number }`.  I added several unit tests to ensure the data was coming back with the `_score` field, and that's working.

Here's a list of other "side quests" that I pushed up to this branch:
1. Updated the `match` syntax to only allow string-based fields, like `name match "(string)"` but not `num match "(string)"`
2. Added a "Scope" concept, and refactored Parameters to add params to the Scope.  This paves the way forward for the parent selector.
3. Added the concept of `score` vs `scoreRaw`.  I want there to be a way to distinguish between type-safe methods vs allowing any raw string, so `scoreRaw` is the version that allows `string`.
4. I applied this same `raw` logic to the other place where we allow raw expressions, so I renamed `filter` to `filterRaw` (and marked `filter` as deprecated)

scottrippey (approved):

PR #340, opened 2025-02-06
Feature: Score: added changeset

scottrippey:
I forgot to add a changeset for #338


soleoak (approved):
👍

scottrippey:
The deprecation of `filter` does make it fit the "minor" description, but being this is so new, I'm going to push as a patch.

Issue #342, opened 2025-02-07
`.field()` seems to return only `never[] | null`


cjcartier:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Bug report

```markdown
I have been playing around with V1 and the implementation of `.field()` seems to be mistyped. Even in the arcade. If I were to go off the arcade, I believe the underlying query still works. But the TS generated from it appears to have some issues. 

Something like 


export const internalLinkFragment = q.fragmentForType().project(sub => ({
  link: sub.field('reference').deref().field('seo.slug', q.slug('slug')),
  blank: q.boolean().optional(),
  alt: q.string().optional(),
}));


returns 


type LinkQuery = {
    link: never[] | null;
    blank: boolean | undefined;
    alt: string | undefined;
}


and the part of the query above `.field('seo.slug', q.slug('slug'))` returns the following ts error: 
`Argument of type '"seo.slug"' is not assignable to parameter of type 'never'`

I even pulled a query straight from the docs: 


const query = q.star.filterByType('product').project(sub => ({
  imageUrl: sub.field('image.asset').deref().field('url', q.string()),
  slug: sub.field('slug.current', q.string()),
  price: sub.field('price', q.number()),
}));

export type QueryType = InferResultItem<typeof query>;
 

QueryType returns 

type something = {
    slug: never[] | null;
    imageUrl: never[] | null;
    price: never[] | null;
}


And has that same error in the same section for `.field('url', q.string())`

`Argument of type '"url"' is not assignable to parameter of type 'never'`
```

cjcartier:
Oh! I also meant to mention that the type always returns an array as well. 

For example
`product: sub.field('product').field('name', q.string())`
and 
`product: sub.field('product[]').field('name', q.string())`

will both return `product: never[] | null`

My hope would be that it would return 
1. `product: string[]`
2. `product: string`

matzexcom:
Something seems wrong with dot notion selects (i.e., `slug.current`, `image.asset`). I tried to avoid it in my migration and use `.field()` every time.
```TS
slug: sub.field('slug').field('current', q.string().nullable()),

// will be
slug: string | null
```

scottrippey:
So a few things are going on here.
[...]

Issue #332, opened 2025-02-03
Dereferencing Sanity `image.asset` field typing problem


matzexcom:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### Bug report

Hi, thanks for your work, I am in the middle of migrating from 0.x to 1.x

I have some problems with the `deref` of the Sanity Image field.
A normal Sanity Image field has the following structure

```
image?: {
    asset?: {
      _ref: string
      _type: 'reference'
      _weak?: boolean
      [internalGroqTypeReferenceTo]?: 'sanity.imageAsset'
    }
    hotspot?: SanityImageHotspot
    crop?: SanityImageCrop
    _type: 'image'
  }

```

My understanding now is that I can do

```
.project((sub) => ({
  imageUrl: sub.field('image.asset').deref().field('url'),
  ...
```

but then typescript complains `Argument of type 'url' is not assignable to parameter of type never` because the `deref` seems to not knowing the result type of `.field('image.asset').deref()`

I tried to use some kind of fragment like 

```
const sanityImage = q.fragmentForType<'sanity.imageAsset'>().project({
  url: q.string().optional().nullable(),
})

.project((sub) => ({
  imageFragment: sub.field('image.asset').deref().project(sanityImage),

```
but than I got the error message from groqd `['⛔️ Error: this projection has an invalid property ⛔️'] | ['⛔️ A parser can only be used for known properties; \'url\' is not known ⛔️']`. The same problem the `defref` does not know the type from `'image.asset' `

I found a work around for that, so that typescript is happy and the `defref` knows that  `'asset'` is a `SanityImageAsset` type.

```
.project((sub) => ({
  image: sub.field('image').project((subsub) => ({
    url: subsub.field('asset').deref().field('url'),
  })),

```
but it seems not intended. Is there some mistake on my side or is that a bug of the `defref` method?

-----

In your docs you often wrote for example

```
q.star.filterByType("product").project(sub => ({
  imageUrl: sub.field("image").deref().field("url", q.string()),
```

but that results in `{"imageUrl": image->.url}` groq query string and that does not work, because the `_ref` is in the `asset` field

scottrippey:
It does seem like this is a bug in `deref`. I will investigate!
Also, you are correct that our doc samples are wrong for images, I'll get that fixed too!

scottrippey:
[...]

PR #344, opened 2025-02-10
Feature: improved projection paths

scottrippey:
- Improves "Projection Paths" implementation, which is used by `.filterBy(...)`, `.field(...)`, and `.project(...)`
- Projection paths now supports deeper selectors, like `"images[].hotspot.x"`
- Projection paths now supports optional fields, like `"image.asset"`
- Projection paths has proper return types for these types, eg. `"images[].hotspot.x": null | Array<null | number>` 

# Testing
All functionality is fully covered by unit tests, please review.

Addresses #332, #342


soleoak (approved):
Looks good to me 🦖.

scottrippey:
> For future reference, is there a way to test this locally by installing the PR version of the package? It would be nice to try to reproduce the issues cited with your PR package changes.

@soleoak There's probably a way to `npm link` this package and test it locally on a project.  However, the best approach for this lib is to test via unit tests.  When someone reports an issue, I try to recreate with a unit test, and usually I'm able to.  Then, once it's fixed, I can feel confident moving forward that I'm safe against regressions!
TL;DR: the unit tests for this project are reliable, extensive, and critical.

packages/groqd/src/commands/projectField.test.ts
```diff
@@ -109,6 +114,20 @@ describe("field (naked projections)", () => {
...
+
+    it("can correctly project deeply optional values", () => {
+      type DeepOptional = {
+        image?: {
+          asset?: {
+            _type: "reference";
+          };
+        };
+      };
+      const query = q.as<DeepOptional>().field("image.asset");
```
scottrippey, line 126: @soleoak Here is where I added a specific unit test for the reported issue.  It failed at first, obviously, and now it verifies that the original problem is solved 🎉

PR #363, opened 2025-02-21
Docs: Prefer separate `zod` imports

scottrippey:


packages/groqd/README.md
```diff
@@ -21,7 +21,7 @@ Read the official documentation at https://commerce.nearform.com/open-source/gro
 `GroqD` uses a strongly-typed chaining syntax to build queries:
 
 ```ts
-import { createGroqBuilder } from 'groqd';
+import { createGroqBuilder, zod } from 'groqd';
```
scottrippey, line 24: 🤔 Would it be better to just drop this, and do:
```
import { z } from 'zod'
```

carbonrobot: I think it should be `z`, since that is what the `zod` docs also use.

scottrippey: My only hesitation to refactor this to `import from 'zod'` is the fact that the `q.default(...)` utility is still relevant.  So I'm not going to make that sweeping change just yet.

I'll update these imports to `z`.

scottrippey:
Merged this PR as part of #365

PR #370, opened 2025-04-07
Allow for marks, on a child, to be undefined when using q.contentBlock()

BrianWalters:
### Description

If, for whatever reason, the marks property of a portable text block is empty or undefined, then the projection made by the q.contentBlock function will not match and the query will fail. I found there are (albeit rare) circumstances where the marks property can be removed or pruned. So in this PR, marks is now optional and will default to an empty array. It will remain as the type `string[]` to users so there are no breaking changes.

Also upgraded groqd-legacy to vitest ^2.1 to ensure both types and assertions are tested.

Fixes # (issue)

#### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### How Has This Been Tested?

Added a test to contentBlock.test.ts.

- [ ] I have included a changeset if this change will require a version change to one of the packages.
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/packages/groqd-legacy/src/contentBlock.test.ts
+++ b/packages/groqd-legacy/src/contentBlock.test.ts
@@ -54,7 +54,7 @@ describe("contentBlock", () => {
     ]);
   });
 
-  it.skip("can handle blocks without marks", async () => {
+  it("can handle blocks without marks", async () => {
     const { data } = await runUserQuery(
       q("*")
         .filterByType("user")
--- a/packages/groqd-legacy/src/contentBlock.test.ts
+++ b/packages/groqd-legacy/src/contentBlock.test.ts
@@ -52,6 +52,25 @@ describe("contentBlock", () => {
       { _type: "link", href: "https://google.com" },
     ]);
   });
+
+  it("can handle blocks without marks", async () => {
+    const { data } = await runUserQuery(
+      q("*")
+        .filterByType("user")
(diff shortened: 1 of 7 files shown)


packages/groqd-legacy/src/contentBlock.test.ts
```diff
@@ -46,12 +46,33 @@ describe("contentBlock", () => {
...
       >();
+    expectTypeOf(data.bio[0].children[0].marks).toEqualTypeOf<string[]>();
     expect(Array.isArray(data.bio)).toBeTruthy();
     expect(data.bio[0]._type === "block").toBeTruthy();
     expect(data.bio[0].markDefs).toEqual([
       { _type: "link", href: "https://google.com" },
     ]);
   });
+
+  it.skip("can handle blocks without marks", async () => {
```
scottrippey:
```suggestion
  it("can handle blocks without marks", async () => {
```

BrianWalters: just fixed

scottrippey (approved):
Looks great, good test coverage 👍

PR #384, opened 2025-05-08
Fixed: conditional validation

scottrippey:
- Adds tests for conditional queries with validation
- Adds support for multiple conditions matching


scottrippey:
Merged as part of #385

